Issue #9, opened 2016-04-02
PR/commit status


jakirkham:
Would be nice to have some sort of PR/commit status (green/red) to tell us if the linting is passing or not as well. This will help us avoid merging bad recipes by accident. Also, it could help contributors see in summary form what is still wrong with a recipe.

jakirkham:
Looks like this might be very [straightforward]( https://github.com/PyGithub/PyGithub/blob/master/github/Commit.py#L154 ).

pelson:
🎉  - great idea @jakirkham!

jakirkham:
Thanks for implementing it.

PR #6, opened 2016-03-31
No recipe

pelson:
This is a testing PR which shouldn't be closed. I encourage everybody to unsubscribe from it, as this is the foundation for some of the testing of this service.

PR #5, opened 2016-03-31
Bad branch

pelson:
This is a testing PR which shouldn't be closed. I encourage everybody to unsubscribe from it, as this is the foundation for some of the testing of this service.

PR #11, opened 2016-04-02
WIP: Adds GitHub status for commits

jakirkham:
Fixes https://github.com/conda-forge/conda-forge-webservices/issues/9

Attempts to add a GitHub status for each commit. Turns out that this is pretty simple to do with GitPython and PyGithub (at least it seems). However, needs to be tested somehow. Also, may need some token permission tweaks. This intended to be extremely basic, but could be built on by others in the future. Some examples below.

* Adding a URL (could point to the posted comment) (tricky as the URL is needed when posting status)
* Could add a pending status. Trick as it might accidentally create two statuses (pending and final).

ref: https://developer.github.com/v3/repos/statuses/
ref: https://github.com/PyGithub/PyGithub/blob/v1.26.0/github/Commit.py#L153
ref: http://gitpython.readthedocs.org/en/stable/tutorial.html#understanding-objects

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -38,6 +38,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
         repo.remotes.origin.fetch('pull/{pr}/head:pr/{pr}'.format(pr=pr_id))
         repo.refs['pr/{}'.format(pr_id)].checkout()
+        sha = repo.head.object.hexsha
         recipes = [y for x in os.walk(tmp_dir)
                    for y in glob(os.path.join(x[0], 'meta.yaml'))]
         all_pass = True
@@ -74,6 +75,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
     {{}}
     """.format(recipe_code_blocks)).format('\n'.join(messages))
 
+    commit = repository.get_commit(sha)
     if not recipe_dirs:
         message = textwrap.dedent("""
             Hi! This is the friendly conda-forge-admin automated user.
@@ -83,8 +85,10 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
             """)
     elif all_pass:
         message = good
+        commit.create_status("success", description="All recipes are excellent.", context="conda-linter-bot")
     else:
         message = bad
+        commit.create_status("failure", description="Some recipes need some changes.", context="conda-linter-bot")
 
     return message
 


conda_forge_webservices/linting.py
```diff
@@ -38,6 +38,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
         repo.remotes.origin.fetch('pull/{pr}/head:pr/{pr}'.format(pr=pr_id))
         repo.refs['pr/{}'.format(pr_id)].checkout()
+        sha = repo.head.object.hexsha
```
jakirkham: Will need a text string (not binary) of the SHA. Hoping that is what this is.

jakirkham: Note we also want the same commit that we generate our comment for. So that is what I am trying to get here.

conda_forge_webservices/linting.py
```diff
@@ -74,6 +75,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
     {{}}
     """.format(recipe_code_blocks)).format('\n'.join(messages))
 
+    commit = repository.get_commit(sha)
```
jakirkham: Gets a PyGithub `Commit` object so that we can add a status.

conda_forge_webservices/linting.py
```diff
@@ -83,8 +85,10 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
             """)
     elif all_pass:
         message = good
+        commit.create_status("success", description="All recipes are excellent.", context="conda-linter-bot")
```
jakirkham: The status strings are pretty specific. So, must be `"success"` here. The rest is just what I filled in. Can be changed.

conda_forge_webservices/linting.py
```diff
@@ -83,8 +85,10 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
     else:
         message = bad
+        commit.create_status("failure", description="Some recipes need some changes.", context="conda-linter-bot")
```
jakirkham: Same story here. Though I think we should use the same `context` in both cases whatever we decide to call it.

jakirkham:
Not sure what is up with the test failure. Seems the `GH_TOKEN` is not set.

pelson:
> Not sure what is up with the test failures. Seems the GH_TOKEN is not set.

Indeed. You need to commit to this repo for the tests to pass (it is pretty intrinsically requiring a GH_TOKEN, so no other option here).

In terms of the approach - I'm keen to keep the "figure out the message" from the "post the message" as it made the whole thing easier to test. Therefore, rather than passing just a string out of the function, we need to pass some context of success, failure etc. and then do the status outside of that function.

:+1: - on the concept of making a status message on each PR

jakirkham:
> Indeed. You need to commit to this repo for the tests to pass (it is pretty intrinsically requiring a GH_TOKEN, so no other option here).

Sorry, not following. Haven't had coffee yet. :coffee: I hope this doesn't mean we need to commit a `GH_TOKEN`. Does this mean commit privileges to this repo?

> In terms of the approach - I'm keen to keep the "figure out the message" from the "post the message" as it made the whole thing easier to test. Therefore, rather than passing just a string out of the function, we need to pass some context of success, failure etc. and then do the status outside of that function.

No complaints with having/keeping comments. Just thinking it would be nice to have statuses, as well. They are a nice simple statistic. Particularly as people start moving large chunks of their personal recipes here.

As far as testing goes, we can actually use the status API to read what the status is for a commit. However, it will likely be difficult for me to test this while I am getting these `GH_TOKEN` issues.

Where would you rather have this status set? They seem to need much of the same information.

pelson:
> Sorry, not following. Haven't had coffee yet. :coffee: I hope this doesn't mean we need to commit a GH_TOKEN. Does this mean commit privileges to this repo?

You would need commit privs to this repo to get the green light (I've added you). Just push to a feature branch and propose another PR and you should get CI passing.

> As far as testing goes, we can actually use the status API to read what the status is for a commit. 

:+1: - we may need to get the PRs #4, #5 and #6 re-opened to set the status (they are closed at the minute as they have diverging histories, but the intention was for them to remain open for testing)

You can also test locally with your own GH_TOKEN:

```
GH_TOKEN=$(~/.conda-smithy/github.token) nosetests conda_forge_webservices
```

jakirkham:
This is being replaced by this PR ( https://github.com/conda-forge/conda-forge-webservices/pull/12 ).

PR #12, opened 2016-04-05
WIP: Adds GitHub status for commits (Pt. 2)

jakirkham:
Fixes https://github.com/conda-forge/conda-forge-webservices/issues/9
Closes https://github.com/conda-forge/conda-forge-webservices/pull/11

Second attempts at adding a GitHub status to each commit. Turns out that this is pretty simple to do with GitPython and PyGithub (at least it seems). However, needs to be tested somehow. Also, may need some token permission tweaks. This intended to be extremely basic, but could be built on by others in the future. Some examples below.

* Adding a URL (could point to the posted comment) (tricky as the URL is needed when posting status)
* Could add a pending status. Trick as it might accidentally create two statuses (pending and final).

ref: https://developer.github.com/v3/repos/statuses/
ref: https://github.com/PyGithub/PyGithub/blob/v1.26.0/github/Commit.py#L153
ref: http://gitpython.readthedocs.org/en/stable/tutorial.html#understanding-objects

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -38,6 +38,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
         repo.remotes.origin.fetch('pull/{pr}/head:pr/{pr}'.format(pr=pr_id))
         repo.refs['pr/{}'.format(pr_id)].checkout()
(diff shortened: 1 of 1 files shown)


jakirkham:
Seems I can't re-open the test PRs that were closed.

jakirkham:
I think there are some issues with my initial implementation. However, it looks like we will need to re-work it anyways based on the feedback I have gotten.

PR #15, opened 2016-04-14
Add status to pull requests

pelson:
Closes https://github.com/conda-forge/conda-forge-webservices/pull/12

Takes #12 further to abstract more information out of the compute_lint_message function.

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -119,6 +119,20 @@ def comment_on_pr(owner, repo_name, pr_id, message):
     return my_last_comment
 
 
+def set_pr_status(owner, repo_name, lint_info):
+    gh = github.Github(os.environ['GH_TOKEN'])
+
+    user = gh.get_user(owner)
+    repo = user.get_repo(repo_name)
+    commit = repo.get_commit(lint_info['sha'])
+    if lint_info['status'] == 'good':
+        commit.create_status("success", description="All recipes are excellent.",
+                             context="conda-linter-bot")
+    else:
+        commit.create_status("failure", description="Some recipes need some changes.",
+                             context="conda-linter-bot")
+
+
 def main():
     import argparse
     parser = argparse.ArgumentParser()
@@ -134,19 +148,7 @@ def main():
 
     if args.enable_commenting:
         comment_on_pr(owner, repo_name, args.pr, lint_info['message'])
-
-        gh = github.Github(os.environ['GH_TOKEN'])
-
-        user = gh.get_user(owner)
-        repo = user.get_repo(repo_name)
-        commit = repo.get_commit(lint_info['sha'])
-        if lint_info['status'] == 'good':
-            commit.create_status("success", description="All recipes are excellent.",
-                                 context="conda-linter-bot")
-        else:
-            commit.create_status("failure", description="Some recipes need some changes.",
-                                 context="conda-linter-bot")
-
+        set_pr_status(owner, repo_name, lint_info)
     else:
         print('Comments not published, but the following would have been the message:\n{}'.format(lint_info['message']))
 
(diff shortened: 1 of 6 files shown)


jakirkham:
Sorry, I haven't played with the status message for awhile. Thanks for giving it a try. If there is anything I can do to help, please let me know.

pelson:
> Sorry, I haven't played with the status message for awhile. Thanks for giving it a try. If there is anything I can do to help, please let me know.

No problems. I'm just trying to pull some of the context out - you were right that all of the bits were conveniently in that function, but I didn't want to pollute its purpose.
Hopefully this change will take it in a healthy direction.

jakirkham:
> No problems. I'm just trying to pull some of the context out - you were right that all of the bits were conveniently in that function, but I didn't want to pollute its purpose.

Yeah, it was a bit tricky, but I think we could figure something out. We just have to balance speed/time with quality here. Also, we can always go back and refactor.

> Hopefully this change will take it in a healthy direction.

Sure, it has tests. So, that is already an improvement on what I did. :smile:

conda_forge_webservices/linting.py
```diff
@@ -1,22 +1,14 @@
-# TODO: Add an interface to do this from the CLI:
-# conda-linting-service conda-forge/staged-recipes 123
-
+from contextlib import contextmanager
+from glob import glob
 import os
-import tornado.escape
-import tornado.httpserver
-import tornado.ioloop
-import tornado.web
```
jakirkham: Were we just never using these?

pelson: Nope. They were split out into the webapp code.

pelson:
So I think this is good to go, but I'm not around now until 12hrs, so will hold fire on deploying this until then.

conda_forge_webservices/linting.py
jakirkham: Ah ok. So these are just leftover imports. Thanks for clarifying.

jakirkham:
Is this failure just caused by having it on your fork? Should we move the PR over to the main repo?

jakirkham:
Here is what I am seeing with the status.

* Show up in the failing case. :+1:
* Missing on the passing. (should be fixed)
* Missing on the no recipe case. (should it show up?)

Seems like the passing case needs some help.

As for the missing case, not sure what to do. Should we pass it and say no recipes found in the summary? Just simply not have a status (as now)? Some other behavior? I think it would be nice to have some status just to verify it is working, but it probably doesn't need to be the same as a good recipe. Though it should not be a failure.

pelson:
> Is this failure just caused by having it on your fork?

Yes. I'm able to run these locally using my own PAT:

```
GH_TOKEN=$(cat ~/.conda-smithy/github.token.pelson) nosetests conda_forge_webservices
```

You should be able to run the tests locally too with your own if you like.

pelson:
Deployed. I've upgraded the dyno to hobby and tested this with https://github.com/conda-forge/staged-recipes/pull/364 (and we are getting a nice status!) 🎉

jakirkham:
Thanks @pelson for bringing this one home. This is truly wonderful.

jakirkham:
So, this one ( https://github.com/conda-forge/staged-recipes/pull/365 ) appears to be passing, but the recipes still have no tests.

pelson:
> So, this one ( conda-forge/staged-recipes#365 ) appears to be passing, but the recipes still have no tests.

Is that just because it hasn't triggered a lint since it was deployed? Looks like the last check was > 7 hours ago...

jakirkham:
Sorry, I was going to mention that soon after it appeared correct. Either I'm not awake :sleeping: or something magical happened :fireworks:. The failure is showing up correctly now.

PR #21, opened 2016-04-19
Added a link to the last comment.

### LINKED ISSUE #20: Add details link for CI status
This might be tricky. However, if we time it right, this should work. Would be nice to have a "Details" link added to the linter bot's CI status that points to the most recent comment. We should be able to do this in cases where the status has changed. To handle cases where it hasn't, we should read the link from the last status before doing anything and then reuse it in the status for the new commit if nothing has changed.

pelson:
Closes #20.

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -119,7 +119,7 @@ def comment_on_pr(owner, repo_name, pr_id, message):
     return my_last_comment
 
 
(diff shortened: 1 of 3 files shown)


pelson:
I've redeployed, and we also have conda-smithy v0.9 🎉

jakirkham:
Wow you are fast. Thanks for getting this one in. Now we can have frosting. :cake:

PR #23, opened 2016-04-20
Fixed failing test for v0.9

pelson:
v0.9 of conda-smithy made the good recipe bad, so I've fixed the recipe, and extended the scope of the bad message.


conda_forge_webservices/tests/linting/test_compute_lint_message.py
```diff
@@ -46,6 +46,7 @@ def test_bad_recipe(self):
          * The summary item is expected in the about section.
          * The recipe could do with some maintainers listed in the "extra/recipe-maintainers" section.
          * The recipe must have some tests.
+         * The recipe must have a `build/number` section.
```
jakirkham: :+1: I think this is especially helpful for users less familiar with the format (e.g. devs on the fringe of Python development or not aware of conda/conda-build greatness). After all we suggest in the Readme to change this number. However, if it is not already there, then we may have some confused users.

PR #33, opened 2016-07-24
Change variable `open` to `is_open`

jakirkham:
We don't want to be overriding keywords like `open`. So, change the variable to a different name. It is unfortunate that Python allows this at all.


jakirkham:
Guessing the failure is caused by this not being on the repo itself.

jakirkham:
Failing because of the lack of the token. However, I think this should work fine. Thoughts on how to proceed?

PR #34, opened 2016-07-24
Add status page

jakirkham:
This requires the `statuspage` package be created. Though it looks like that recipe should be ready to go. It may be worth a final review as much has changed since the last one. See PR ( https://github.com/conda-forge/staged-recipes/pull/775 ).

Using `statuspage` is really simple. We just point it at repo, give it a token, and tell it to run `update`. This is basically what `status.py` below is. Right now we are using `GH_TOKEN` as the environment variable. However, I don't really like this as I don't want the status page to potentially get throttled by the linter's rate limit. I already have another bot setup ( @conda-forge-status ) for this task. So we should discuss how we want to do this.

The rest of the code in `webapp.py` is just trying to register us for issue update events on the status page. We may find this is too frequent and need to tone it down. However, I don't have a good feel for that right now without see this in action. So, it may just be easiest to start with this and optimize later. Please review this part carefully.

### DIFF
--- a/conda_forge_webservices/status.py
+++ b/conda_forge_webservices/status.py
@@ -1,0 +1,32 @@
+import os
+import subprocess
+
+
+def update(token=None):
+    if token is None:
+        token = os.environ["GH_TOKEN"]
+
+    subprocess.check_call([
+        "statuspage",
+        "update",
+        "--org",
+        "conda-forge",
+        "--name",
+        "status",
+        "--token",
+        token
+    ])
+
+def main():
+    import argparse
+
+    parser = argparse.ArgumentParser(
+        prog="status",
+        description="Updates the status page.",
+    )
+
+    update(os.environ["GH_TOKEN"])
+
+
+if __name__ == '__main__':
+    main()
--- a/conda_forge_webservices/status.py
+++ b/conda_forge_webservices/status.py
@@ -4,7 +4,7 @@ import subprocess
 
 def update(token=None):
     if token is None:
-        token = os.environ["GH_TOKEN"]
+        token = os.environ["STATUS_GH_TOKEN"]
 
     subprocess.check_call([
         "statuspage",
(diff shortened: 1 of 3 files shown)


jakirkham:
So there actually is a `statuspage` package. If you could look over these changes @pelson, I would be very appreciative.

pelson:
At some point we are likely to want to separate the linting endpoint from the status endpoint. For that reason, I recommend that we add a new handler and new endpoint in the map (https://github.com/jakirkham/conda-forge-webservices/blob/e8191f5d636b3ce3a8dde82cba92d2f0bfc876bf/conda_forge_webservices/webapp.py#L100).

Hitting something like ```status/hook``` would make a lot of sense.

jakirkham:
SGTM

jakirkham:
Alright, I think this is what you want to see. Though I'm a bit confused as to how the feedstocks get registered. So I'm unclear as to whether I have properly limited notifications to the status page repo beforehand.

conda_forge_webservices/status.py
```diff
@@ -0,0 +1,32 @@
...
+
+def main():
+    import argparse
+
+    parser = argparse.ArgumentParser(
+        prog="status",
+        description="Updates the status page.",
+    )
+
+    update(os.environ["GH_TOKEN"])
```
jakirkham: Forgot to rename this one. Fixing in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/35 ).

jakirkham:
FYI enabled `push` for the webhook as well. This is important if we change the layout of the status page in its templates. That way the page will get regenerated if a PR is merged for instance.

jakirkham:
Also, gave `Core` write permissions to that repo. Not sure that admin permissions are really required for regular usage. Though we can certainly discuss.

jakirkham:
Enabled `Issue comment` for the webhook too as it appears the comments will show up in a log for the current problem on the webpage.

jakirkham:
However, we didn't handle these events initially. So we need to update the script to handle them. Fortunately that is pretty easy. See PR ( https://github.com/conda-forge/conda-forge-webservices/pull/37 ) for these changes.

PR #37, opened 2016-08-18
Handle some other status page events

jakirkham:
There are a few other events that appear to result in the status page updating. One of them is if there is a comment on an issue. In this case, the comments will show up under the active issue in the status page. The other is if a commit is pushed to the repo. So if someone updates the logo or the templates, it should change what the webpage looks like. In both cases, we do the exact same thing as if an issue is opened/closed. As a result we just do an update for all of these cases.

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -91,12 +91,12 @@ class StatusHookHandler(tornado.web.RequestHandler):
 
         if event == 'ping':
             self.write('pong')
-        elif event == 'issues':
+        elif event == 'issues' or event == 'issue_comment' or event == 'push':
             body = tornado.escape.json_decode(self.request.body)
             repo_name = body['repository']['name']
             owner = body['repository']['owner']['login']
 
-            # Only do something if it is an issue on the status page
+            # Only do something if it involves the status page
             if owner == 'conda-forge' and repo_name == 'status':
                 status.update()
         else:


pelson:
I'm wondering if we need to start considering debouncing some of the hooks so that we don't produce many identical updates to the status repo. Additionally, we are going to need to be careful if we are getting a hook back after each commit - because we are adding commits whenever we receive a hook we can end up in an ugly infinite loop 😞

jakirkham:
Was just about to clarify on that point.

`statuspage` won't actually generate a new commit if nothing changed. This is thanks to a fix in [0.4.0]( https://github.com/jayfk/statuspage/blame/master/CHANGELOG.md#L15-L16 ). In any event, we still get an extra request, but should get no more than one.

So the workflow looks like this now if we start with a non-commit change.

1. Do something with issue.
2. Webhook to Heroku.
3. statuspage generates commit.
4. Webhook to Heroku.
5. Nothing changed end.

If the change is a commit we pushed to `master` (i.e. merged PR), what happens in 1 changes, but the other steps remain the same.

jakirkham:
Though interestingly it seems we are getting an error on that additional run (step 5). Before we got a 404 error understandably as we [raised it]( https://github.com/conda-forge/conda-forge-webservices/blob/beb4b83154ef7566bd3f22988a42604fe9d979fa/conda_forge_webservices/webapp.py#L103-L105 ).

With this change in, we are now getting a 500 error. So maybe something is going wrong when handling the no commit case. Perhaps it is giving a non-zero exit code and we need to suppress that somehow.

jakirkham:
Looking at the Heroku logs it seems we are assuming an entry is defined that is not. A short snippet is shown below.

```
File "/app/conda_forge_webservices/webapp.py", line 97, in post
KeyError: 'login'
```

jakirkham:
This PR ( https://github.com/conda-forge/conda-forge-webservices/pull/39 ) should fix it.

PR #39, opened 2016-08-19
Fix push event handling

jakirkham:
Handle the slightly different response for a `push` event. It uses `name` instead of `login` used by `issues` or `issue_comment` events. This is a minor tweak to the changes presented in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/37 ).

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -93,11 +93,10 @@ class StatusHookHandler(tornado.web.RequestHandler):
             self.write('pong')
         elif event == 'issues' or event == 'issue_comment' or event == 'push':
             body = tornado.escape.json_decode(self.request.body)
-            repo_name = body['repository']['name']
-            owner = body['repository']['owner']['login']
+            repo_full_name = body['repository']['full_name']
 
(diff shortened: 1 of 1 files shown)


pelson:
Definitely needs a comment (and an xref back to here) I think.

jakirkham:
They all have access to the repo's fullname (in other words slug). So, I have changed it to that for simplicity. Honestly I doubt this check is really necessary (just like the owner check for the linter) as we have already limited this by adding the webhook where we saw fit to. In any event, having the check doesn't do any harm so I have left it.

jakirkham:
Does this seem sensible now?

jakirkham:
Hmm...seems to have not rebuilt. Thought it was suppose to do that on `git push`. 😕

The webhook doesn't have an error FWICT. Retriggering it didn't seem to make a difference.

jakirkham:
Looks like it has now been rebuilt. Not sure what was going on before.

jakirkham:
Yay! The webhook is now passing. 😄 Also it successfully did nothing on `push` as the page was already up-to-date.

PR #40, opened 2016-08-19
Use Python 3

jakirkham:
This is as much a question as an actual request. Is there a good reason not to use Python 3 (or for use of Python 2)? As Python 3 handles unicode much better, this may already help solve some annoying linter issues. Thus am proposing we use Python 3 for testing, building, and deployement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ sudo: false
 
 env:
   matrix:
-    - python=2.7
-#    - python=3.5
+#    - python=2.7
+    - python=3.5
 
   global:
     # Adds a GH_TOKEN (pelson) for testing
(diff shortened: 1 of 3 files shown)


pelson:
There was a reason, but I can't remember what it was, and whether it is still valid.
Would you mind testing by pushing the branch onto this repo? The CI should then turn green.

jakirkham:
Sure, let's see what happens. Pushed to `use_python_3` on the repo.

pelson:
Easy to revert if this goes badly. Would you mind deleting that branch now @jakirkham?

jakirkham:
Sure. Removed.

Initially this wasn't working for the status page in a brief test today. Based on the logs it looked like Heroku was forcing ANSI encoding and that was causing some problems. So I set `LANG=en_US.UTF-8` in the config vars and that seemed to fix it. Just something to be aware of.

Tested the linter as well with PR ( https://github.com/conda-forge/rank_filter-feedstock/pull/12 ). That seems to be working alright too.

Issue #45, opened 2016-08-27
Ignore already merged recipes at staged-recipes


jakirkham:
It would be good to have the linter use the git trick that the CIs use at staged-recipes to ignore already merged recipes in PRs. This would make sure that it only lints the recipe in each PR and does not waste time or confuse users by linting recipes that are waiting for feedstock conversion. See PR ( https://github.com/conda-forge/staged-recipes/pull/1373 ) that added this functionality to staged-recipes.

jakirkham:
PR ( https://github.com/conda-forge/conda-forge-webservices/pull/53 ) has a very nice implementation for this problem. In fact, I may try to go back and improve on the work in PR ( https://github.com/conda-forge/staged-recipes/pull/1373 ) based on these proposed changes.

PR #47, opened 2016-08-31
WIP: Purges merged recipes from PRs for linting

jakirkham:
Fixes https://github.com/conda-forge/conda-forge-webservices/issues/45

I'm trying to decide how much value this has. Right now we are grabbing the PR's `head` ref. As such, it is not already merged with `master`. This differs from what the CIs generally due, which is checkout the PR's `merge` ref. As a result, recipes are not really going to *sneak* into one's PR. They will definitely be visible locally in one's recipe directory. Though maybe this will help avoid the impulse of deleting them, which can make the diffs noise. It should also be more flexible as we add more examples or rename existing ones.

xref: https://github.com/conda-forge/staged-recipes/pull/1373

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -17,6 +17,9 @@ def tmp_directory():
     yield tmp_dir
     shutil.rmtree(tmp_dir)
 
+def find_recipes(a_dir):
+    return [y for x in os.walk(a_dir)
+            for y in glob(os.path.join(x[0], 'meta.yaml'))]
 
 def compute_lint_message(repo_owner, repo_name, pr_id, clear_history=None):
     gh = github.Github(os.environ['GH_TOKEN'])
@@ -33,26 +36,22 @@ def compute_lint_message(repo_owner, repo_name, pr_id, clear_history=None):
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
 
         # Find the lingering recipes.
+        historic_recipes = []
         if clear_history:
             repo.refs[clear_history].checkout()
-            staged_recipes = set(map(lambda _: _.name, repo.tree()['recipes']))
+            historic_recipes = find_recipes(tmp_dir)
 
         # Checkout the PR head and get the list of recipes.
         repo.remotes.origin.fetch('pull/{pr}/head:pr/{pr}'.format(pr=pr_id))
         repo.refs['pr/{}'.format(pr_id)].checkout()
         sha = str(repo.head.object.hexsha)
-        recipes = [y for x in os.walk(tmp_dir)
-                   for y in glob(os.path.join(x[0], 'meta.yaml'))]
+        recipes = find_recipes(tmp_dir)
         all_pass = True
         messages = []
 
         # Exclude the recipes already merged in `master`.
+        recipes = list(set(recipes) - set(historic_recipes))
         recipe_dirs = map(os.path.dirname, recipes)
-        if clear_history:
-            recipe_dirs = dict(map(lambda _: (os.path.basename(_), _), recipe_dirs))
-            recipe_dirs = [
-                recipe_dirs[_] for _ in (set(recipe_dirs.keys()) - staged_recipes)
-            ]
 
         # Sort the recipes for consistent linting order (which glob doesn't give us).
         recipe_dirs = sorted(recipe_dirs)
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -28,6 +28,8 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
 
     with tmp_directory() as tmp_dir:
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
+
+        # Checkout the PR head and get the list of recipes.
         repo.remotes.origin.fetch('pull/{pr}/head:pr/{pr}'.format(pr=pr_id))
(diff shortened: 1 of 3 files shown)


conda_forge_webservices/linting.py
```diff
@@ -28,15 +28,25 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
 
     with tmp_directory() as tmp_dir:
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
+
+        # Find the lingering recipes.
+        staged_recipes = []
+        if repo_owner == 'conda-forge' and repo_name == 'staged-recipes':
+            r.refs['master'].checkout()
```
jakirkham: This is probably not necessary honestly. Just seems like a good safety measure.

conda_forge_webservices/linting.py
```diff
@@ -28,15 +28,25 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
         sha = str(repo.head.object.hexsha)
         recipes = [y for x in os.walk(tmp_dir)
                    for y in glob(os.path.join(x[0], 'meta.yaml'))]
         all_pass = True
         messages = []
+
+        # Exclude the recipes already merged in `master`.
         recipe_dirs = [os.path.dirname(recipe) for recipe in recipes
-                       if os.path.basename(os.path.dirname(recipe)) != 'example']
+                       if os.path.basename(os.path.dirname(recipe)) not in staged_recipes]
```
jakirkham: We can probably do a better comparison here. May try to add such a thing to this PR.

conda_forge_webservices/linting.py
```diff
@@ -28,15 +28,28 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
         messages = []
-        recipe_dirs = [os.path.dirname(recipe) for recipe in recipes
-                       if os.path.basename(os.path.dirname(recipe)) != 'example']
+
+        # Exclude the recipes already merged in `master`.
+        recipe_dirs = map(os.path.dirname, recipes)
+        recipe_dirs = dict(map(lambda _: (os.path.basename(_), _), recipe_dirs))
+        recipe_dirs = [
+                recipe_dirs[_] for _ in (set(recipe_dirs.keys()) - staged_recipes)
+        ]
```
jakirkham: Tried to make things a little more optimal by doing them in stages (for reuse) and working with sets (for difference). Not sure if it helps much, but I imagine this code gets hit a lot. So anything we can do helps.

jakirkham: Also did a very basic test locally and it seemed ok.

conda_forge_webservices/linting.py
```diff
@@ -28,15 +28,28 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
 
     with tmp_directory() as tmp_dir:
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
+
+        # Find the lingering recipes.
+        staged_recipes = set()
+        if repo_owner == 'conda-forge' and repo_name == 'staged-recipes':
```
jakirkham: This sort of makes testing hard. Suggestions welcome.

jakirkham:
Added an optional argument to `compute_lint_message` that normally doesn't do anything. However that should allow us control whether this cleaning behavior occurs or not.

jakirkham:
I'd like to setup another test case, but I don't know how. I have added it as `ok_recipe` in the same named branch.

conda_forge_webservices/linting.py
```diff
@@ -136,16 +151,24 @@ def set_pr_status(owner, repo_name, lint_info, target_url=None):
 
 def main():
     import argparse
+
     parser = argparse.ArgumentParser()
     parser.add_argument('repo')
     parser.add_argument('pr', type=int)
     parser.add_argument('--enable-commenting', help='Turn on PR commenting',
                         action='store_true')
+    parser.add_argument('--clear-history', nargs='?', type=str, default=argparse.SUPPRESS,
```
pelson: How about calling this `ignore-existing-recipes`?

pelson: `action="store_true"` is always a winner 😉

conda_forge_webservices/linting.py
```diff
@@ -17,26 +17,41 @@ def tmp_directory():
...
+    if clear_history is True:
+        clear_history = 'master'
+
     with tmp_directory() as tmp_dir:
         repo = Repo.clone_from(repo.clone_url, tmp_dir)
+
+        # Find the lingering recipes.
+        historic_recipes = []
+        if clear_history:
+            repo.refs[clear_history].checkout()
```
pelson: I wonder if we can just use what was checked out. It should be the default branch as we have just cloned it?

conda_forge_webservices/linting.py
jakirkham: Thought about that, but I wanted the option to specify which branch potentially.

conda_forge_webservices/linting.py
jakirkham: The main reason I went this route instead of just `master` or leaving it alone was so we could get some test coverage. After all if we want to work these test PRs `master` might not be an option if we want to see this do something non-trivial.

jakirkham:
Replacing with PR ( https://github.com/conda-forge/conda-forge-webservices/pull/53 ) as that feels like a much cleaner approach. Thanks for the suggestions @pelson. They have been incorporated over there.

PR #50, opened 2016-09-01
Refactor `find_recipes` in linting

jakirkham:
Refactors out the function `find_recipes` in linting, which should make it easier to reuse in different phases.

PR #53, opened 2016-09-02
Purges recipes in base of PRs before linting

jakirkham:
Fixes https://github.com/conda-forge/conda-forge-webservices/issues/45
Closes https://github.com/conda-forge/conda-forge-webservices/pull/47

Initially I had written PR ( https://github.com/conda-forge/conda-forge-webservices/pull/47 ) and it quickly got wildly unmanageable by attempting to handle all the use cases of the linter (staged-recipes and feedstocks) as well as providing a testable interface. Ultimately looking at it filled me with such a level of disgust 😝 that I had to just rewrite this from scratch.

In this PR, we take advantage of remote refs in the same way the linter already does, but we go one step further. Instead of simply getting the PR `head` ref, we also get the PR `merge` ref. The latter is provided in particular for use by CIs so that they can checkout the PR as if it were already merged into the base branch. In the event that we do want to remove recipes that are in the base branch (now referred to as `base`, we can now determine what the latest commit in the `base` is (pre-merge). As there should only be two parent commits of the merge, there should be no confusion as to which one is in `base`. If there is some confusion about what `base` is, we should be sure to raise an error explain what that confusion is.

However, there are some cases we do need to be aware of that are problematic. One is the case where the `merge` ref is not defined. For now this is limited to the case of a PR that was created in conflict. The other is the case where the PR has come to be in conflict. In this latter case, the `merge` ref may be defined (if the PR was ever mergeable), but would be out-of-date. The first case could easily be solved by trying to retrieve it and then erroring out. The second case however cannot be easily caught via `git` as the `merge` ref is still present. As a result, we simply query the GitHub API if the PR is mergeable. This way we can be sure to cover both cases without issue. Since we like to provide comments for the user it seems nice to do so in this case. As this seemed like useful functionality on its own, it has been broken out into a separate PR ( https://github.com/conda-forge/conda-forge-webservices/pull/55 ) and merged back into here. For more details on this point please read issue ( https://github.com/isaacs/github/issues/757 ).

Once `base` has been determined, it is a simple matter to get the recipes from it and exclude those recipes from the list we find in `head`. The result being the list proposed in this PR.

After writing this, I noticed that there were a bunch of small useful usable pieces that could be broken out and addressed separately even if this attempt doesn't take. These include PRs ( https://github.com/conda-forge/conda-forge-webservices/pull/49 ), ( https://github.com/conda-forge/conda-forge-webservices/pull/50 ), ( https://github.com/conda-forge/conda-forge-webservices/pull/51 ), and ( https://github.com/conda-forge/conda-forge-webservices/pull/55 ). Hopefully those can be merged independently of this.

Have tried to set up PR ( https://github.com/conda-forge/conda-forge-webservices/pull/54 ) for testing purging. Merge conflict issues are tested in PRs ( https://github.com/conda-forge/conda-forge-webservices/pull/56 ) and ( https://github.com/conda-forge/conda-forge-webservices/pull/57 ). Also have written a few tests for this functionality. Hopefully we can get a good idea of whether this works or not. If you are subscribed to that testing PR, please unsubscribe yourself as it will be very noisy.

xref: https://github.com/conda-forge/staged-recipes/pull/1373

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -22,7 +22,7 @@ def find_recipes(a_dir):
     return [os.path.dirname(y) for x in os.walk(a_dir)
             for y in glob(os.path.join(x[0], 'meta.yaml'))]
 
-def compute_lint_message(repo_owner, repo_name, pr_id):
+def compute_lint_message(repo_owner, repo_name, pr_id, ignore_base=False):
     gh = github.Github(os.environ['GH_TOKEN'])
 
     owner = gh.get_user(repo_owner)
@@ -74,6 +74,19 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
 
             return lint_info
 
+        # Collect recipes from base that should be ignored.
+        base_recipes = []
+        if ignore_base:
+            num_parents = len(ref_merge.commit.parents)
+            assert num_parents == 2, textwrap.dedent("""
+                   Expected merging our PR with the base branch would have two parents.
+                   Instead there were %i parents found. :/
+                   """ % num_parents)
+            base_commit = (set(ref_merge.commit.parents) - {ref_head.commit}).pop()
+            ref_base = repo.create_head("pull/{pr}/base".format(pr=pr_id), base_commit)
+            ref_base.checkout()
+            base_recipes = find_recipes(tmp_dir)
+
         # Get the list of recipes and prep for linting.
         ref_merge.checkout()
         recipes = find_recipes(tmp_dir)
@@ -81,16 +94,13 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
         messages = []
 
         # Exclude some things from our list of recipes.
-        recipe_dirs = [recipe for recipe in recipes
-                       if os.path.basename(recipe) != 'example']
-
         # Sort the recipes for consistent linting order (which glob doesn't give us).
-        recipe_dirs = sorted(recipe_dirs)
+        pr_recipes = sorted(set(recipes) - set(base_recipes))
 
-        rel_recipe_dirs = []
-        for recipe_dir in recipe_dirs:
+        rel_pr_recipes = []
+        for recipe_dir in pr_recipes:
             rel_path = os.path.relpath(recipe_dir, tmp_dir)
-            rel_recipe_dirs.append(rel_path)
+            rel_pr_recipes.append(rel_path)
             try:
                 lints = conda_smithy.lint_recipe.main(recipe_dir)
             except Exception as err:
@@ -102,7 +112,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
                                                              '\n'.join(' * {}'.format(lint) for lint in lints)))
 
     # Put the recipes in the form "```recipe/a```, ```recipe/b```".
-    recipe_code_blocks = ', '.join('```{}```'.format(r) for r in rel_recipe_dirs)
+    recipe_code_blocks = ', '.join('```{}```'.format(r) for r in rel_pr_recipes)
 
     good = textwrap.dedent("""
     Hi! This is the friendly automated conda-forge-linting service.
@@ -121,7 +131,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
     {{}}
     """.format(recipe_code_blocks)).format('\n'.join(messages))
 
-    if not recipe_dirs:
+    if not pr_recipes:
         message = textwrap.dedent("""
             Hi! This is the friendly automated conda-forge-linting service.
             
@@ -192,11 +202,14 @@ def main():
     parser.add_argument('pr', type=int)
     parser.add_argument('--enable-commenting', help='Turn on PR commenting',
                         action='store_true')
+    parser.add_argument('--ignore-base',
+                        help='Ignore recipes in the base branch of the PR',
+                        action='store_true')
 
     args = parser.parse_args()
     owner, repo_name = args.repo.split('/')
 
-    lint_info = compute_lint_message(owner, repo_name, args.pr)
+    lint_info = compute_lint_message(owner, repo_name, args.pr, args.ignore_base)
 
     if not lint_info:
         print('Linting was skipped.')
--- a/conda_forge_webservices/tests/linting/test_main.py
+++ b/conda_forge_webservices/tests/linting/test_main.py
@@ -14,6 +14,22 @@ class TestCLI_recipe_lint(unittest.TestCase):
         out, _ = child.communicate()
         self.assertEqual(child.returncode, 0, out)
 
+    def test_cli_success_ok_above_ignored_good(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '54',
+                                  '--enable-commenting', '--ignore-base'],
+                                 stdout=subprocess.PIPE, env=os.environ)
+        out, _ = child.communicate()
+        self.assertEqual(child.returncode, 0, out)
+
+    def test_cli_success_ok_beside_ignored_good(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '62',
+                                  '--enable-commenting', '--ignore-base'],
+                                 stdout=subprocess.PIPE, env=os.environ)
+        out, _ = child.communicate()
+        self.assertEqual(child.returncode, 0, out)
+
     def test_cli_success_conflict_ok(self):
         child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
                                   'conda-forge/conda-forge-webservices', '56', '--enable-commenting'],
--- a/conda_forge_webservices/tests/test_webapp.py
+++ b/conda_forge_webservices/tests/test_webapp.py
@@ -49,3 +49,28 @@ class TestBucketHandler(TestHandlerBase):
                                               {'message': mock.sentinel.message},
                                               target_url=mock.sentinel.html_url)
 
+    @mock.patch('conda_forge_webservices.linting.compute_lint_message', return_value={'message': mock.sentinel.message})
+    @mock.patch('conda_forge_webservices.linting.comment_on_pr', return_value=mock.MagicMock(html_url=mock.sentinel.html_url))
+    @mock.patch('conda_forge_webservices.linting.set_pr_status')
+    def test_staged_recipes(self, set_pr_status, comment_on_pr, compute_lint_message):
+        PR_number = 16
+        body = {'repository': {'name': 'staged-recipes',
+                               'clone_url': 'repo_clone_url',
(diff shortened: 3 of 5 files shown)


conda_forge_webservices/linting.py
```diff
@@ -6,45 +6,93 @@
...
+                Please ping the 'conda-forge/core' team (using the @ notation in a comment) if you believe this is a bug.
+                """)
+            status = 'merge conflict'
+            sha = str(repo.head.object.hexsha)
+
+            lint_info = {'message': message,
+                         'status': status,
+                         'sha': sha}
+
+            return lint_info
```
jakirkham: This part catches a merge conflict. However, it only captures a merge conflict when the PR is opened. See issue ( https://github.com/isaacs/github/issues/757 ) for a longer explanation of why.

We could add a GitHub API check for conflicts too. The message wouldn't necessarily be so useful in that case though not necessarily a problem either. Mainly it would make sure we aren't handling duplicates incorrectly.

conda_forge_webservices/linting.py
```diff
@@ -6,45 +6,98 @@
...
+        if ignore_base:
+            assert len(ref_merge.commit.parents) != 0, textwrap.dedent("""\
+                   Merging our PR with the base branch results in an orphaned commit. :/
+                   """)
+            assert len(ref_merge.commit.parents) != 1, textwrap.dedent("""\
+                   Merging our PR with the base branch results in a commit with one parent. :/
+                   """)
+            assert len(ref_merge.commit.parents) == 2, textwrap.dedent("""\
+                   Merging our PR with the base branch results in more than two parents. :/
+                   """)
```
jakirkham: In reality, we can probably cut these down to one informative assertion. Not sure that this problem ever occurs. Still I'd like the info if it ever does.

conda_forge_webservices/tests/linting/test_main.py
```diff
@@ -14,6 +14,28 @@ def test_cli_success_bad(self):
         out, _ = child.communicate()
         self.assertEqual(child.returncode, 0, out)
 
+    def test_cli_success_ok(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '54',
+                                  '--enable-commenting', '--ignore-base'],
+                                 stdout=subprocess.PIPE, env=os.environ)
+        out, _ = child.communicate()
+        self.assertEqual(child.returncode, 0, out)
```
jakirkham: Only tested the ignore base case with CLI and two recipes. After all, we already test the two recipe case with comments either with or without ignoring base. So testing with two recipes where we don't ignore the base isn't really so different from the CLI perspective. Though I could be talked into adding it if people see value in it.

conda_forge_webservices/tests/test_webapp.py
```diff
@@ -40,7 +40,7 @@ def test_good_header(self, set_pr_status, comment_on_pr, compute_lint_message):
 
         self.assertEqual(response.code, 200)
         compute_lint_message.assert_called_once_with('conda-forge', 'repo_name',
-                                                     PR_number)
+                                                     PR_number, False)
```
jakirkham: As we now have a default value of `False`, we need to update this test to acknowledge it is there.

conda_forge_webservices/webapp.py
```diff
@@ -75,7 +75,8 @@ def post(self):
 
             # Only do anything if we are working with conda-forge, and an open PR.
             if is_open and owner == 'conda-forge':
-                lint_info = linting.compute_lint_message(owner, repo_name, pr_id)
+                lint_info = linting.compute_lint_message(owner, repo_name, pr_id,
+                                                         repo_name == 'staged-recipes')
```
jakirkham: This is the magic touch. We only ignore the base branch at `conda-forge/staged-recipes`. Otherwise we always include the base branch (i.e. at feedstocks).

conda_forge_webservices/linting.py
```diff
@@ -6,45 +6,98 @@
...
+        try:
+            repo.remotes.origin.fetch([
+                'pull/{pr}/head:pull/{pr}/head'.format(pr=pr_id),
+                'pull/{pr}/merge:pull/{pr}/merge'.format(pr=pr_id)
+            ])
+            ref_head = repo.refs['pull/{pr}/head'.format(pr=pr_id)]
+            ref_merge = repo.refs['pull/{pr}/merge'.format(pr=pr_id)]
+        except GitCommandError:
+            # Either `merge` doesn't exist because the PR was opened
+            # in conflict or it is closed and it can't be the latter.
```
jakirkham: Sometimes the `merge` ref isn't defined.

First case is when the PR is closed (it is still defined, though incorrectly). This case should be handled already by this [line]( https://github.com/conda-forge/conda-forge-webservices/pull/53/files#diff-924207d49c849574f35d69090b77eea3R77 ) (only running on open PRs).

Second case is when the PR has always had merge conflicts since it was opened. If the conflicts were ever resolved in the PR, then this is no longer a concern. This is a case that we do have to handle. Currently we don't do anything about it here, and instead check on this [line]( https://github.com/conda-forge/conda-forge-webservices/pull/53/files#diff-c238032bbeab767fb95f0ccd3d23585eR53 ) for merge conflicts. This seems to work well.

conda_forge_webservices/tests/test_webapp.py
```diff
@@ -49,3 +49,28 @@ def test_good_header(self, set_pr_status, comment_on_pr, compute_lint_message):
...
+                'pull_request': {'number': PR_number,
+                                 'state': 'open'}}
+
+        response = self.fetch('/conda-linting/hook', method='POST',
+                              body=json.dumps(body),
+                              headers={'X-GitHub-Event': 'pull_request'})
+
+        self.assertEqual(response.code, 200)
+        compute_lint_message.assert_called_once_with('conda-forge', 'staged-recipes',
+                                                     PR_number, True)
```
jakirkham: Test running on staged-recipes to verify it will ignore the base branch of PRs.

conda_forge_webservices/linting.py
jakirkham: Shortened.

pelson:
@jakirkham - looks like this overlaps quite significantly with https://github.com/conda-forge/conda-forge-webservices/pull/55. Would you mind revitalizing this PR, and I'll happily merge.

jakirkham:
Yeah was trying to break this into smaller chunks as it started to seem there was other nice pieces in it. Have cleaned that content out of here and merged back in with `master`. Everything passes.

conda_forge_webservices/linting.py
```diff
@@ -52,23 +67,33 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
+                   """ % num_parents)
+            base_commit = (set(ref_merge.commit.parents) - {ref_head.commit}).pop()
+            ref_base = repo.create_head("pull/{pr}/base".format(pr=pr_id), base_commit)
+            ref_base.checkout()
+            base_recipes = find_recipes(tmp_dir)
+
         # Get the list of recipes and prep for linting.
-        recipes = [y for x in os.walk(tmp_dir)
-                   for y in glob(os.path.join(x[0], 'meta.yaml'))]
+        ref_merge.checkout()
```
jakirkham: I've switched to checking out the merge commit as this matches closer to what CIs do normally. In the end everything works the same as verified by the passing tests.

jakirkham:
Anything else needed from me, @pelson?

jakirkham:
Cleared out some of the smaller PRs that were attached to this change in the hopes of making the review (and merge 😁) easier.

conda_forge_webservices/linting.py
jakirkham: TL;DR Sometimes there is no merge ref basically must be a merge conflict. This is already caught with PR ( https://github.com/conda-forge/conda-forge-webservices/pull/55 ). So we don't need to do anything here.

conda_forge_webservices/linting.py
```diff
@@ -55,22 +67,33 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
+        if ignore_base:
+            num_parents = len(ref_merge.commit.parents)
+            assert num_parents == 2, textwrap.dedent("""
+                   Expected merging our PR with the base branch would have two parents.
+                   Instead there were %i parents found. :/
+                   """ % num_parents)
+            base_commit = (set(ref_merge.commit.parents) - {ref_head.commit}).pop()
+            ref_base = repo.create_head("pull/{pr}/base".format(pr=pr_id), base_commit)
+            ref_base.checkout()
+            base_recipes = find_recipes(tmp_dir)
```
jakirkham: This is the magic sauce. Find the recipes in the base branch and make a list (if we are to ignore them).

conda_forge_webservices/linting.py
```diff
@@ -55,22 +67,33 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
         all_pass = True
         messages = []
 
         # Exclude some things from our list of recipes.
-        recipe_dirs = [recipe for recipe in recipes
-                       if os.path.basename(recipe) != 'example']
-
         # Sort the recipes for consistent linting order (which glob doesn't give us).
-        recipe_dirs = sorted(recipe_dirs)
+        pr_recipes = sorted(set(recipes) - set(base_recipes))
```
jakirkham: Now we just remove the base branch recipes from our list (assuming we were asked to look for any 😉).

jakirkham:
Also tried to draw attention to the key parts to hopefully make it clearer.

jakirkham:
Broke out another piece of the functionality here in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/61 ) and merged it back in here. That piece should be pretty easily mergeable, which should help us clean up this PR a bit more.

conda_forge_webservices/linting.py
jakirkham: Also this code has been broke out as PR ( https://github.com/conda-forge/conda-forge-webservices/pull/60 ).

jakirkham:
For reference, not having this functionality is causing confusion and unneeded issues. Please see PR ( https://github.com/conda-forge/staged-recipes/pull/1395 ) for an example.

jakirkham:
So it seems like we are both happy with this and ready to merge. As not having it is causing some unneeded confusion/problems and you seem very busy of late @pelson, I'm going to move towards putting this in myself. Hope this is ok. If you do find there are things needing fixes, I'm more than happy to look at them and address them as needed.

jakirkham:
Starting with merging PR ( https://github.com/conda-forge/conda-forge-webservices/pull/61 ), which is part of this so as to clean this PR up and make it more clear what it is doing.

jakirkham:
PR ( https://github.com/conda-forge/conda-forge-webservices/pull/61 ) has been merged. Have rebased this on `master` to clear that part out.

jakirkham:
This is now passing and cleaner. Merging this as well. Again if there are issues, please let me know and I will be happy to address them.

jakirkham:
See this [comment]( https://github.com/conda-forge/staged-recipes/pull/1534#issuecomment-248810585 ) for a live field test and explanation.

PR #55, opened 2016-09-02
Catch and raise merge conflicts

jakirkham:
Closes https://github.com/conda-forge/conda-forge-webservices/pull/51

This checks to see if a PR will have merge conflicts and raises an error if it does. This should help the occasional confused or git tormented person find their way back to a mergeable PR. It also saves us from linting things when they are in a non-functioning state. After all we would not otherwise know there was a merge conflict. Testing occurs in PRs ( https://github.com/conda-forge/conda-forge-webservices/pull/56 ) and ( https://github.com/conda-forge/conda-forge-webservices/pull/57 ).

### DIFF
--- a/conda_forge_webservices/tests/linting/test_main.py
+++ b/conda_forge_webservices/tests/linting/test_main.py
@@ -14,6 +14,20 @@ class TestCLI_recipe_lint(unittest.TestCase):
         out, _ = child.communicate()
         self.assertEqual(child.returncode, 0, out)
 
+    def test_cli_success_conflict_ok(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '56', '--enable-commenting'],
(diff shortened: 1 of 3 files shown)


jakirkham:
Looks like this needs some tweaks. Please see PR ( https://github.com/conda-forge/conda-forge-webservices/pull/68 ).

conda_forge_webservices/tests/linting/test_main.py
```diff
@@ -14,6 +14,20 @@ def test_cli_success_bad(self):
         out, _ = child.communicate()
         self.assertEqual(child.returncode, 0, out)
 
+    def test_cli_success_conflict_ok(self):
```
pelson: In general please try to avoid these kind of tests where possible - they should be used *only* if we are unable to ensure coverage through unit testing.

No action needed on this.

jakirkham: By which you are meaning CLI tests? The main reason I added these was to ensure some uncaught exception wouldn't propagate out. Though maybe we have already covered that case well enough.

Issue #46, opened 2016-08-27
Respect skip in commits


jakirkham:
Would be nice if we could skip the linter in commits. This is particularly useful for maintenance PRs at staged-recipes. Would propose the following be supported.

* `[ci skip]`
* `[skip ci]`
* `[lint skip]`
* `[skip lint]`

jakirkham:
Please see PR ( https://github.com/conda-forge/conda-forge-webservices/pull/60 ).

PR #60, opened 2016-09-12
Skip linting via commit message

jakirkham:
Closes https://github.com/conda-forge/conda-forge-webservices/issues/46

Adds the ability to skip the linter if needed. Useful if the change doesn't affect the linter like maintenance or doc changes at staged-recipes. Could also include re-rendering PRs potentially. Looks for 4 possible things in the head commit of the PR:

* `[ci skip]`
* `[skip ci]`
* `[lint skip]`
* `[skip lint]`

If any of these 4 is found, then it won't lint that PR. Tests are added to cover all of these cases.

Testing All CIs PR: https://github.com/conda-forge/conda-forge-webservices/pull/58 https://github.com/conda-forge/conda-forge-webservices/pull/65
Testing Linter only PR: https://github.com/conda-forge/conda-forge-webservices/pull/59 https://github.com/conda-forge/conda-forge-webservices/pull/66

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -56,6 +56,18 @@ def compute_lint_message(repo_owner, repo_name, pr_id, ignore_base=False):
             ref_head = repo.refs['pull/{pr}/head'.format(pr=pr_id)]
         sha = str(ref_head.commit.hexsha)
 
+        # Check if the linter is skipped via the commit message.
+        skip_msgs = [
+            "[ci skip]",
+            "[skip ci]",
+            "[lint skip]",
+            "[skip lint]",
+        ]
+        commit_msg = repo.commit(sha).message
+        should_skip = any([msg in commit_msg for msg in skip_msgs])
+        if should_skip:
+            return {}
+
         # Raise an error if the PR is not mergeable.
         if not mergeable:
             message = textwrap.dedent("""
--- a/conda_forge_webservices/tests/linting/test_compute_lint_message.py
+++ b/conda_forge_webservices/tests/linting/test_compute_lint_message.py
@@ -19,6 +19,22 @@ def tmp_directory():
 
 
 class Test_compute_lint_message(unittest.TestCase):
+    def test_skip_ci_recipe(self):
+        lint = compute_lint_message('conda-forge', 'conda-forge-webservices', 58)
+        self.assertFalse(lint)
+
+    def test_skip_lint_recipe(self):
+        lint = compute_lint_message('conda-forge', 'conda-forge-webservices', 59)
+        self.assertFalse(lint)
+
+    def test_ci_skip_recipe(self):
+        lint = compute_lint_message('conda-forge', 'conda-forge-webservices', 65)
+        self.assertFalse(lint)
+
+    def test_lint_skip_recipe(self):
+        lint = compute_lint_message('conda-forge', 'conda-forge-webservices', 66)
+        self.assertFalse(lint)
+
     def test_good_recipe(self):
         expected_message = textwrap.dedent("""
         Hi! This is the friendly automated conda-forge-linting service.
--- a/conda_forge_webservices/tests/linting/test_main.py
+++ b/conda_forge_webservices/tests/linting/test_main.py
@@ -7,6 +7,34 @@ import unittest
 
 
 class TestCLI_recipe_lint(unittest.TestCase):
+    def test_cli_skip_ci(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '58', '--enable-commenting'],
+                                 stdout=subprocess.PIPE, env=os.environ)
+        out, _ = child.communicate()
+        self.assertEqual(child.returncode, 0, out)
+
+    def test_cli_skip_lint(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '59', '--enable-commenting'],
+                                 stdout=subprocess.PIPE, env=os.environ)
+        out, _ = child.communicate()
+        self.assertEqual(child.returncode, 0, out)
+
+    def test_cli_ci_skip(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '65', '--enable-commenting'],
+                                 stdout=subprocess.PIPE, env=os.environ)
+        out, _ = child.communicate()
+        self.assertEqual(child.returncode, 0, out)
+
+    def test_cli_lint_skip(self):
+        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
+                                  'conda-forge/conda-forge-webservices', '66', '--enable-commenting'],
+                                 stdout=subprocess.PIPE, env=os.environ)
+        out, _ = child.communicate()
+        self.assertEqual(child.returncode, 0, out)
+
     def test_cli_success_bad(self):
         child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
                                   'conda-forge/conda-forge-webservices', '17', '--enable-commenting'],
--- a/conda_forge_webservices/tests/linting/test_main.py
+++ b/conda_forge_webservices/tests/linting/test_main.py
@@ -14,27 +14,6 @@ class TestCLI_recipe_lint(unittest.TestCase):
         out, _ = child.communicate()
         self.assertEqual(child.returncode, 0, out)
 
-    def test_cli_skip_lint(self):
-        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
-                                  'conda-forge/conda-forge-webservices', '59', '--enable-commenting'],
-                                 stdout=subprocess.PIPE, env=os.environ)
-        out, _ = child.communicate()
-        self.assertEqual(child.returncode, 0, out)
-
-    def test_cli_ci_skip(self):
-        child = subprocess.Popen([sys.executable, '-m' 'conda_forge_webservices.linting',
-                                  'conda-forge/conda-forge-webservices', '65', '--enable-commenting'],
-                                 stdout=subprocess.PIPE, env=os.environ)
-        out, _ = child.communicate()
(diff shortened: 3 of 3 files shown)


conda_forge_webservices/linting.py
```diff
@@ -37,6 +37,18 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
         sha = str(repo.head.object.hexsha)
 
+        # Check if we should use CI.
+        skip_msgs = [
+            "[ci skip]",
+            "[skip ci]",
+            "[lint skip]",
+            "[skip lint]",
+        ]
+        commit_msg = repo.head.object.message
```
jakirkham: Will need to fix this when PR ( https://github.com/conda-forge/conda-forge-webservices/pull/53 ) is merged as this will be pointing to `master` instead of the PR `head` ref.

jakirkham: Fixed. Now we get the commit message associated with the `sha`. This works the same before and after those PRs are merged.

conda_forge_webservices/linting.py
```diff
@@ -37,6 +37,18 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
         repo.refs['pr/{}'.format(pr_id)].checkout()
         sha = str(repo.head.object.hexsha)
 
+        # Check if the linter is skipped via the commit message.
+        skip_msgs = [
+            "[ci skip]",
+            "[skip ci]",
+            "[lint skip]",
+            "[skip lint]",
```
jakirkham: Thought about `linter`, but `lint` is a bit more compact and still gets the same idea across. Could always add `linter` later.

conda_forge_webservices/linting.py
```diff
@@ -173,7 +185,9 @@ def main():
 
     lint_info = compute_lint_message(owner, repo_name, args.pr)
 
-    if args.enable_commenting:
+    if not lint_info:
+        print('Linting was skipped.')
```
jakirkham: Really no preference what goes here. This was just something that seemed reasonable.

jakirkham:
Can you please take a look at this one when you have a chance, @pelson?

jakirkham:
Updated to include tests for all 4 cases. Noted this in the OP.

conda_forge_webservices/linting.py
```diff
@@ -56,6 +56,18 @@ def compute_lint_message(repo_owner, repo_name, pr_id, ignore_base=False):
...
 
+        # Check if the linter is skipped via the commit message.
+        skip_msgs = [
+            "[ci skip]",
+            "[skip ci]",
+            "[lint skip]",
+            "[skip lint]",
+        ]
+        commit_msg = repo.commit(sha).message
+        should_skip = any([_ in commit_msg for _ in skip_msgs])
```
pelson: We shouldn't use `_` as a variable that we actually care about. `message` / `msg` would be fine IMO.

conda_forge_webservices/linting.py
```diff
@@ -56,6 +56,18 @@ def compute_lint_message(repo_owner, repo_name, pr_id, ignore_base=False):
...
+        if should_skip:
+            return {}
```
pelson: Doesn't this just mean that there was no lint? If so, wouldn't we just get a message saying everything was fine?

conda_forge_webservices/tests/linting/test_compute_lint_message.py
```diff
@@ -19,6 +19,22 @@ def tmp_directory():
 
 
 class Test_compute_lint_message(unittest.TestCase):
+    def test_skip_ci_recipe(self):
```
pelson: Please avoid too many of these types of tests - they are hard to maintain and encourage us to avoid seeking tighter unit testing.

jakirkham: Sorry what is wrong with these? They are testing the `compute_lint_message` function. There isn't a smaller unit to test. What would you we do propose instead?

conda_forge_webservices/linting.py
jakirkham: Only picked it as we iterate with it, but am happy to rename it. 👍

conda_forge_webservices/linting.py
jakirkham: Nope. Sadly I had to break out part of this and get it in to fix a bug with a PR ( https://github.com/conda-forge/conda-forge-webservices/pull/55 ). So, I'll have to refer you to PR ( https://github.com/conda-forge/conda-forge-webservices/pull/69 ) for details. The short answer is if we have a trivial `dict` we do nothing.

conda_forge_webservices/linting.py
jakirkham: Done.

jakirkham:
So am a little confused as to what to do regarding testing as I do have tests for `compute_lint_message` function, which is the smallest exposed API, in the same style as other tests here. While I'm not against us exploring improving the test suite, I'm still not clear on what this entails. Perhaps we should discuss this in a new issue ( https://github.com/conda-forge/conda-forge-webservices/issues/72 ) to iron out what it entails and work on a likely well-deserved separate PR to address that issue.

conda_forge_webservices/tests/linting/test_compute_lint_message.py
jakirkham: Were you meaning to refer to the `subprocess.Popen` tests below?

pelson: Yes. Thanks @jakirkham.

pelson:
> So am a little confused as to what to do regarding testing as I do have tests for compute_lint_message function, which is the smallest exposed API, in the same style as other tests here. 

Sorry for the confusion - I did mean the Popen tests that require a real PR on this repo. The other tests are good IMO.

conda_forge_webservices/tests/linting/test_main.py
```diff
@@ -7,6 +7,34 @@
 
 
 class TestCLI_recipe_lint(unittest.TestCase):
+    def test_cli_skip_ci(self):
```
pelson: If you literally reverted the change to this file (and closed the PRs that go with it), I'm 👍  on merging this PR.

jakirkham:
So the integration tests are testing something important. Please see this [line]( https://github.com/conda-forge/conda-forge-webservices/raw/bdd024603a69ffce623878795e2a7e2a763d6c92/conda_forge_webservices/linting.py#L20 ). Basically they are ensuring a skipped lint doesn't result in an error. Before it would have caused a `KeyError` to be raised in the webservice. Now it doesn't and these tests help verify that.

conda_forge_webservices/tests/linting/test_main.py
jakirkham: Please see the comment below.

conda_forge_webservices/linting.py
```diff
@@ -56,6 +56,18 @@ def compute_lint_message(repo_owner, repo_name, pr_id, ignore_base=False):
...
+        skip_msgs = [
+            "[ci skip]",
+            "[skip ci]",
+            "[lint skip]",
+            "[skip lint]",
+        ]
+        commit_msg = repo.commit(sha).message
+        should_skip = any([msg in commit_msg for msg in skip_msgs])
+        if should_skip:
+            return {}
```
pelson: I've been looking at this for the last few minutes and trying to figure out how I feel about returning an empty result indicating a skip. There are two options I would probably have chosen from:

 * Poison the function by raising a custom SkipLint error
 * Return a correct status dictionary that put the status as "skipped" and the message as nothing.

I think that would reduce the cognitive burden of lines like https://github.com/conda-forge/conda-forge-webservices/pull/69/files#diff-c238032bbeab767fb95f0ccd3d23585eR177.

conda_forge_webservices/tests/linting/test_main.py
pelson: I take your point. We will only need one test to check that line though 😉

jakirkham: Ok, that's fine with me. Unless you have a preference, let's do the one that skips all CIs so it should be quieter.

conda_forge_webservices/linting.py
jakirkham: I see your point and agree we should address it somehow. Though I think we need to give this more thought.

The empty dictionary can be returned in a few cases not related to an intentional skip. It can be returned if a [PR is closed mid-lint]( https://github.com/conda-forge/conda-forge-webservices/pull/68/files#diff-c238032bbeab767fb95f0ccd3d23585eR35 ). There is also the possibility that we end up wanting to limit this [loop]( https://github.com/conda-forge/conda-forge-webservices/pull/68/files#diff-c238032bbeab767fb95f0ccd3d23585eR32 ) to a fixed number of checks, at which point we also need to bail out.

Do we want to handle these all the same way or a different way? This may be worth discussing in a new issue that is assigned to me. Once we have some consensus, I can take it on in a subsequent PR.

<hr>
That all being said, here are my two concerns.

* Reducing cognitive load (in agreement with you).
* Avoiding a noisy linter.

While point 2 sounds good as it is similar to other statuses, it is unclear that it will reduce the cognitive load that much and it is possible we accidentally fall into a noisy linter. So to me 2 is out.

Point 1 does not fit as nicely into our categories. However, it should reduce cognitive load and if we fail to handle it the linter doesn't do anything.

jakirkham:
So I dropped the extra integration tests and just kept one.

However, we need to keep the PRs open. Otherwise we are not testing that skip message was found and the lint skipped. Instead we are testing that we did not lint the closed PR. Please see this [line]( https://github.com/conda-forge/conda-forge-webservices/pull/68/files#diff-c238032bbeab767fb95f0ccd3d23585eR35 ) to better understand what I mean.

pelson:
Thanks @jakirkham. Yep, though I don't like it, you are right. We need to pull the lint computation out of that function so that we can unit test it appropriately. I'm happy to do that as a follow-up activity.

Thanks for all the info. 👍

jakirkham:
Thanks @pelson. I'm also happy to work on this as well. Not trying to put extra work in your lap.

Maybe we should take a look at what others are doing to have proper test suites with GitHub and mimic them here. Have collected some useful references along these lines in issue ( https://github.com/conda-forge/conda-forge-webservices/issues/72 ).

jakirkham:
Also an example of this behavior working correctly can be seen in PR ( https://github.com/conda-forge/staged-recipes/pull/1675 ).

PR #61, opened 2016-09-12
Lint PR merge commit

jakirkham:
Switches to linting the merge commit of PRs instead of the head commit. This is inline with what many other CIs are doing (like Travis CI and AppVeyor). The main reason to do this is to ensure we are not getting false positives from the linter. For example, imagine our PR has an error due to some part of the recipe that is unchanged, but the base branch contains a correction to this issue. We don't want this linting error as it is a false positive and will go away with a simple merge into the base branch.

It is possible to do this now as we correctly raise merge conflicts as an error thanks to PR ( https://github.com/conda-forge/conda-forge-webservices/pull/55 ). The functionality presented here is a smaller piece of the functionality already in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/53 ) to optionally ignore recipes from the base branch. A test is provided to ensure we are getting the intended behavior by using PR ( https://github.com/conda-forge/conda-forge-webservices/pull/54 ).

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -7,7 +7,7 @@ import textwrap
 import time
 
 import requests
-from git import Repo
+from git import GitCommandError, Repo
 import github
 import conda_smithy.lint_recipe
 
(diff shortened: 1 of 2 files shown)


conda_forge_webservices/linting.py
```diff
@@ -32,10 +32,22 @@ def compute_lint_message(repo_owner, repo_name, pr_id):
...
+        try:
+            repo.remotes.origin.fetch([
+                'pull/{pr}/head:pull/{pr}/head'.format(pr=pr_id),
+                'pull/{pr}/merge:pull/{pr}/merge'.format(pr=pr_id)
+            ])
+            ref_head = repo.refs['pull/{pr}/head'.format(pr=pr_id)]
+            ref_merge = repo.refs['pull/{pr}/merge'.format(pr=pr_id)]
+        except GitCommandError:
+            # Either `merge` doesn't exist because the PR was opened
+            # in conflict or it is closed and it can't be the latter.
```
jakirkham: Sometimes the `merge` ref isn't defined.

First case is when the PR is closed. This case should be handled already by this [line]( https://github.com/conda-forge/conda-forge-webservices/pull/53/files#diff-924207d49c849574f35d69090b77eea3R77 ) (only running on open PRs).

Second case is when the PR has always had merge conflicts since it was opened. If the conflicts were ever resolved in the PR, then this is no longer a concern. However if merge conflicts enter the PR, `merge` ref is still defined (though incorrectly). This is a case that we do have to handle.

Currently we don't do anything about these here, and instead check on this [line]( https://github.com/conda-forge/conda-forge-webservices/pull/53/files#diff-c238032bbeab767fb95f0ccd3d23585eR53 ) for merge conflicts. This seems to work well.

jakirkham: TL;DR Sometimes there is no merge ref basically must be a merge conflict. This is already caught with PR ( https://github.com/conda-forge/conda-forge-webservices/pull/55 ). So we don't need to do anything here.

jakirkham:
Also can you please take a look at this one when you have a chance, @pelson?

jakirkham:
Merging so as to clear out a piece of PR ( https://github.com/conda-forge/conda-forge-webservices/pull/53 ) and make that one clearer.

PR #69, opened 2016-09-19
Handle an empty lint

jakirkham:
Broken out from PR ( https://github.com/conda-forge/conda-forge-webservices/pull/60 ).

There are some cases where we don't know whether there will be a need to lint until we have started to compute the lint. Examples include commit messages with notes to skip ci explicitly and PRs closed during linting. To handle the case of an empty lint, we updated the code and tests to check to make sure the lint did happen before proceeding further. This will help prepare for the possibility of a lint being skipped while in progress.

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -151,13 +151,14 @@ def set_pr_status(owner, repo_name, lint_info, target_url=None):
 
     user = gh.get_user(owner)
     repo = user.get_repo(repo_name)
-    commit = repo.get_commit(lint_info['sha'])
-    if lint_info['status'] == 'good':
-        commit.create_status("success", description="All recipes are excellent.",
-                             context="conda-forge-linter", target_url=target_url)
-    else:
-        commit.create_status("failure", description="Some recipes need some changes.",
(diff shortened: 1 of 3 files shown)


conda_forge_webservices/linting.py
```diff
@@ -173,7 +174,9 @@ def main():
 
     lint_info = compute_lint_message(owner, repo_name, args.pr)
 
-    if args.enable_commenting:
+    if not lint_info:
+        print('Linting was skipped.')
```
jakirkham: Really no preference what goes here. This was just something that seemed reasonable.

jakirkham:
Going to go ahead and merge this as it has no actionable effect. It merely proofs against a potentially empty lint.

jakirkham:
So the integration tests are testing something relevant. Please see this [line]( https://github.com/conda-forge/conda-forge-webservices/raw/bdd024603a69ffce623878795e2a7e2a763d6c92/conda_forge_webservices/linting.py#L20 ). Basically they are ensuring a skipped lint doesn't result in an error. Before it would have caused a `KeyError` to be raised in the webservice. Now it doesn't and these tests help verify that.

PR #76, opened 2016-10-08
Made use of the multiprocess capability of Tornado & Heroku.

pelson:


pelson:
Going to take this for a spin to see if we can reduce the response time when concurrent requests come in. I will be keeping an eye on whether we loose log fidelity (I suspect we might) and then we will have to make a call as to whether we revert the change.

jakirkham:
This hasn't been as reliable as we would like. Seems we get a lot internal errors from the linter. I think we might be better off switching to gunicorn. Doesn't look to challenging to [do]( https://groups.google.com/d/msg/python-tornado/tTzZvHaLfb0/CUIaarZAGHIJ ).

PR #89, opened 2016-12-09
Checkout with force during linting

jakirkham:
In some cases (e.g. formatting errors), it may be necessary to checkout with force so as to ignore any changes to the repo since cloning it. As the linter shouldn't make any changes itself or propose any, this should not cause any problems.


jakirkham:
Tested on PR ( https://github.com/conda-forge/glpk-feedstock/pull/18 ) where previously it was unable to lint due to formatting issues. It now works correctly.

Issue #63, opened 2016-09-13
Team update/feedstocks update webservice


jakirkham:
While we have done a good job at limited the effect of these scripts, there are still problems. If they haven't surfaced or resurfaced yet, they will soon. So we should start planning to handle them. Here are some issues that come to mind.
1. Scales poorly
   1. GitHub Rate limit issues.
      1. https://github.com/conda-forge/conda-forge.github.io/issues/88
      2. https://github.com/conda-forge/conda-forge.github.io/issues/172
2. Fragile
   1. Feedstock level issues break it.
      1. https://github.com/conda-forge/conda-forge.github.io/issues/86
      2. https://github.com/conda-forge/conda-forge.github.io/issues/139
      3. https://github.com/conda-forge/conda-forge.github.io/issues/228
3. Inefficient
   1. Disproportional time spent on unchanged feedstocks. ( https://github.com/conda-forge/conda-forge.github.io/issues/88#issuecomment-217548992 )
   2. Run time increased from 2min to 12min in last 6mos.
4. Lagged response
   1. User are frequently surprised they aren't added right away.
      1.1. Recent example ( https://github.com/conda-forge/traitlets-feedstock/pull/5#issuecomment-245930774 ), but there are many.
5. Duplicated functionality.
   1. staged-recipes - https://github.com/conda-forge/staged-recipes/pull/733
   2. webpage repo - https://github.com/conda-forge/conda-forge.github.io/pull/15
6. Poor separation of concerns
   1. Run in CI on webpage repo.
   2. Run on PR merges that consist of...
      1. Minutes
      2. Pinnings
      3. Assorted helper scripts
7. Lacking test coverage

Looking at these issues they are all directly or indirectly because we have been using a global update script. As a result, we are spending lots of time doing work that is not particularly useful and paying a high price for it. However, when going back through the feedstocks repo (which is updated roughly once daily), it looks like ~20 feedstocks are added/changed daily. So there is no reason it should be so difficult to keep up with this especially as often as people want us to for cheap. What can we do?

---

We need to switch from global update to asynchronous local updates. This model has been highly successful for us with things like the linter. To solve this particular problem, we would need a webservice that listens to [`PushEvents`](https://developer.github.com/v3/activity/events/types/#pushevent) on each feedstock. When there is a push event, we do the following.
1. New repos: `conda-smithy` adds webhook for this service (just like the linter).
2. Update the sha of the feedstock in the feedstocks repo.
[...]

Issue #73, opened 2016-09-26
Automated Welcome Message


jakirkham:
@jakirkham commented on [Fri May 20 2016](https://github.com/conda-forge/conda-forge.github.io/issues/149)

This is an idea that @ericdill had (or at least similar to it). I used to do these things by hand, but with the growth of conda-forge. It is a bit hard to stay on top of this and I don't think I am the only one that feels that way. Would be nice to have a bot message that welcomes new people added as maintainers to conda-forge when a PR at staged-recipes is merged. The bot would need to explain the following things:
1. Briefly what [`conda`](http://conda.pydata.org/) is.
2. Briefly what [conda-forge](http://conda-forge.github.io/) is.
3. Briefly what a feedstocks is and which ones will be created.
4. Point out an invite to conda-forge is being sent to them and what that entails.
5. Explain how to download packages from [`conda-forge`](http://anaconda.org/conda-forge/).
6. Explain how to make changes to their feedstocks (fork, branch in fork, PR to main).

While 1 may sound silly, there have been a few people that have joined that never heard of `conda`. So, explaining what it is helps. Similarly 2 is nice as people may have been asked to help maintain, but may only now be figuring out what conda-forge is. Guessing 3-6 don't need justification, but could be wrong. Thoughts would be appreciated.

Note: There was another idea about inviting devs of a package. Unfortunately, that problem can be a bit ill-defined at times (e.g. source not on GitHub). Still that might be worth thinking about.

---

@jakirkham commented on [Tue Sep 13 2016](https://github.com/conda-forge/conda-forge.github.io/issues/149#issuecomment-246705763)

This should be more easily achievable if we convert the team update script to a webservice as described in issue ( https://github.com/conda-forge/conda-forge-webservices/issues/63 ).

---

@jakirkham commented on [Tue Sep 20 2016](https://github.com/conda-forge/conda-forge.github.io/issues/149#issuecomment-247866869)

Template message below.

> Soon you should get an email that invite you to join conda-forge, <GitHub handles>. Once accepted you will be added to a team with the same name as this recipe. Those will give you permissions on the feedstock (repo) for this recipe.
> 
> Make sure when proposing any change that you go through the typical GitHub workflow of forking the feedstock and making changes in your fork that you PR back. Once merged CIs will build and deploy any changes you make.
> 
> Please let us know if you have any questions and welcome to conda-forge. :)

---

[...]

PR #96, opened 2017-05-30
New script to update teams

isuruf:
Partially addresses ( https://github.com/conda-forge/conda-forge-webservices/issues/63 ).

@jakirkham, this script needs a organization hook for push events. Whenever the master branch is updated in a feedstock, this will check the team and update it.

### DIFF
--- a/conda_forge_webservices/update_teams.py
+++ b/conda_forge_webservices/update_teams.py
@@ -2,7 +2,7 @@ from git import GitCommandError, Repo
 from conda_build.metadata import MetaData
 import github
 import os
-
+import subprocess
 from .utils import tmp_directory
 
 def update_team(org_name, repo_name):
@@ -15,24 +15,7 @@ def update_team(org_name, repo_name):
 
     with tmp_directory() as tmp_dir:
         repo = Repo.clone_from(remote_repo.clone_url, tmp_dir)
-        meta = MetaData(tmp_dir)
-        updated_maintainers = set(meta.meta.get('extra', {}).get('recipe-maintainers', []))
-
-        team_name = repo_name[:-len("-feedstock")]
-        team = next(team for team in org.get_teams() if team.name == team_name)
-        current_maintainers = set([maintainer.login.lower() for maintainer in team.get_members()])
-
-        for new_maintainer in updated_maintainers - current_maintainers:
-            headers, data = team._requester.requestJsonAndCheck(
-                "PUT",
-                team.url + "/memberships/" + new_maintainer
-            )
-
-        for old_maintainer in current_maintainers - updated_maintainers:
-            headers, data = team._requester.requestJsonAndCheck(
-                "DELETE",
-                team.url + "/memberships/" + old_maintainer
-            )
+        subprocess.check_output(["conda", "smithy", "register-github", "--organization", org_name, "--add-teams", tmp_dir])
 
 
 if __name__ == '__main__':
--- a/conda_forge_webservices/update_teams.py
+++ b/conda_forge_webservices/update_teams.py
@@ -1,0 +1,44 @@
+from git import GitCommandError, Repo
+from conda_build.metadata import MetaData
+import github
+import os
+
+from .utils import tmp_directory
+
+def update_team(org_name, repo_name):
+    if not repo_name.endswith("-feedstock"):
+        return
+
+    gh = github.Github(os.environ['GH_TOKEN'])
+    org = gh.get_organization(org_name)
+    remote_repo = org.get_repo(repo_name)
+
+    with tmp_directory() as tmp_dir:
+        repo = Repo.clone_from(remote_repo.clone_url, tmp_dir)
+        meta = MetaData(tmp_dir)
+        updated_maintainers = set(meta.meta.get('extra', {}).get('recipe-maintainers', []))
+
+        team_name = repo_name[:-len("-feedstock")]
+        team = next(team for team in org.get_teams() if team.name == team_name)
+        current_maintainers = set([maintainer.login.lower() for maintainer in team.get_members()])
+
+        for new_maintainer in updated_maintainers - current_maintainers:
+            headers, data = team._requester.requestJsonAndCheck(
+                "PUT",
+                team.url + "/memberships/" + new_maintainer
+            )
+
+        for old_maintainer in current_maintainers - updated_maintainers:
+            headers, data = team._requester.requestJsonAndCheck(
+                "DELETE",
+                team.url + "/memberships/" + old_maintainer
+            )
+
+
+if __name__ == '__main__':
+    import argparse
+    parser = argparse.ArgumentParser()
+    parser.add_argument('org')
+    parser.add_argument('repo')
+    args = parser.parse_args()
+    update_team(args.org, args.repo)
--- a/conda_forge_webservices/utils.py
+++ b/conda_forge_webservices/utils.py
@@ -1,0 +1,9 @@
+import shutil
+import tempfile
+from contextlib import contextmanager
+
+@contextmanager
+def tmp_directory():
+    tmp_dir = tempfile.mkdtemp('_recipe')
+    yield tmp_dir
+    shutil.rmtree(tmp_dir)
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -18,6 +18,7 @@ from contextlib import contextmanager
 
 import conda_forge_webservices.linting as linting
 import conda_forge_webservices.status as status
+import conda_forge_webservices.update_teams as update_teams
 
 
 class RegisterHandler(tornado.web.RequestHandler):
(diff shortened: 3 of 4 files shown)


conda_forge_webservices/update_teams.py
```diff
@@ -0,0 +1,44 @@
+from git import GitCommandError, Repo
+from conda_build.metadata import MetaData
+import github
+import os
+
+from .utils import tmp_directory
+
+def update_team(org_name, repo_name):
+    if not repo_name.endswith("-feedstock"):
+        return
```
jakirkham: We should print some sort of message here. Maybe raise an error?

isuruf: We should ignore repos like `conda-forge/staged-recipes`

conda_forge_webservices/webapp.py
```diff
@@ -106,10 +107,32 @@ def post(self):
...
+
+        if event == 'ping':
+            self.write('pong')
+        elif event == 'push':
+            body = tornado.escape.json_decode(self.request.body)
+            repo_name = body['repository']['name']
+            owner = body['repository']['owner']['login']
+            ref = body['ref']
+            # Only do anything if we are working with conda-forge, and a push to master.
+            if owner == 'conda-forge' and ref == "refs/heads/master":
```
jakirkham: I don't think we should restrict this to `master`. We have multibranch feedstocks too. Usually created to handle different versions or variants of a package. If someone adds themselves to a non-`master` branch because they are interested primarily in that version or variant, we should still add them as maintainers.

conda_forge_webservices/webapp.py
```diff
@@ -106,10 +107,32 @@ def post(self):
...
+                update_teams.update_team(owner, repo_name)
```
jakirkham: We might want to do some exception handling here. Potentially including the error message in the status.

conda_forge_webservices/utils.py
```diff
@@ -0,0 +1,9 @@
+import shutil
+import tempfile
+from contextlib import contextmanager
+
+@contextmanager
+def tmp_directory():
+    tmp_dir = tempfile.mkdtemp('_recipe')
+    yield tmp_dir
+    shutil.rmtree(tmp_dir)
```
jakirkham: It might be worth breaking out this change into a smaller PR, which can basically be merged as is.

jakirkham:
Thanks for writing this @isuruf. Added some initial comments above. I think we can reuse much of this for updating the feedstocks repo as well.

jakirkham:
Also added a reference in the OP above.

jakirkham:
As for handling the hooks, I think this is something `conda-smithy` should do during CI registration (or maybe it should be called service registration). This is what we do to add the linter hook currently.

isuruf:
Currently following happens,
1. Teams are created in staged-recipes
2. Teams are updated in conda-forge.github.io
3. conda-smithy also has capability to create teams, but it is not used in 1.

How about something like below?
4. conda-smithy gets a new command update team `conda-smithy update-team`. This would create a team if not found and update the team if found.
5. `conda-smithy update-team` is called in staged-recipes
6. `conda-smithy update-team` is called in conda-forge-webservices.

jakirkham:
That was the plan. :) One can already do this by running `conda smithy register-github --organization conda-forge --add-teams foo` (with `foo` as the feedstock repo). Would note that running `conda smithy register-github` repeatedly is idempotent.

isuruf:
Didn't get the last part. I can just call `conda smithy register-github --organization conda-forge --add-teams foo` after cloning the repo here right?

jakirkham:
That's right. Though it would probably be `foo-feedstock` then (as it is the directory name).

isuruf:
One thing concerns me though.
In order to get the team from github we are getting all the teams and checking for the team name. Since github is sending a paginated list, this makes lots of calls to the github API and we may get rate-limited.

I asked github support whether they have any API to get the team info from the team name, but they replied that they don't. Have you requested for this functionality before? If not, I'll request them to provide this functionality.

isuruf:
@jakirkham, please see my last comment

jakirkham:
Thanks @isuruf.

> Since github is sending a paginated list, this makes lots of calls to the github API and we may get rate-limited.

How many calls do you think this is making? I think this is something that we should be able to estimate. If it is only a couple, I'm not too worried. If it is in the 100s, then we should give this some more thought. We can always cache the listing on Heroku if we can determine it will be very problematic.

> I asked github support whether they have any API to get the team info from the team name, but they replied that they don't. Have you requested for this functionality before? If not, I'll request them to provide this functionality.

I don't think so. In any event, I see no harm in potentially asking them again. Feel free to cc me on the email if you would like.

isuruf:
> How many calls do you think this is making?

It's around 85 (2532 teams/30) calls which means max 58 pushes per hour, which is a lot. For normal operations this is okay, but if somebody does a bulk update, then it'll be a problem. Since conda-forge is expanding, the number 85 is going to increase and at one point, we'll realize that the bot is not working and spend hours or days trying to figure out what is causing the issue.

> We can always cache the listing on Heroku if we can determine it will be very problematic.

Yes, I think this is the best way forward.

> I don't think so. In any event, I see no harm in potentially asking them again. Feel free to cc me on the email if you would like.

I'll send them an email again.

isuruf:
Looking at the script again, it uses (<number-of-teams-in-conda-forge> + <number-of-members-in-all-members>)/30 = 100 github api calls.

jakirkham:
I think we should revisit this again. The Travis CI cron job is bumping up against the time limit. So we need some way to workaround this constraint.

While I get that even this strategy can have issues in the future, I think the days of having one long running script over all feedstocks is over.

isuruf:
With https://github.com/conda-forge/conda-smithy/pull/522, number of github calls is constant.

jakirkham:
Is that PR still WIP?

isuruf:
It is ready now

jakirkham:
Should we upgrade `conda-smithy` here too?

conda_forge_webservices/update_teams.py
jakirkham: Though we can accomplish that by simply not adding a webhook to staged-recipes, right?

jakirkham:
We also will need to add the webhook to all existing feedstocks. Perhaps we can reuse something similar to your token strategy to deploy these changes? Also are new feedstocks already being created with this webhook?

conda_forge_webservices/update_teams.py
isuruf: I originally thought of adding an organization hook, but we can do individual hooks now with conda-smithy

jakirkham:
Also had pushed this to the `isuruf/updateteams` branch on this repo.

isuruf:
> Perhaps we can reuse something similar to your token strategy to deploy these changes?

Yes, we could

> Also are new feedstocks already being created with this webhook?

yes they are. I can use one of them to test.

jakirkham:
Wondering whether we should fuse in a solution to issue ( https://github.com/conda-forge/conda-forge-webservices/issues/73 ) as well. Namely we would leave a comment on the commit that we checked welcoming new team members. This would be useful for two reasons. First we can see that team members that should be added did get added. Second the new team members are notified that they have been added (potentially with details about how to accept the GitHub invite to the org if necessary).

isuruf:
One issue is that, this checks pushes to master and not PR, so how can we do that?
I've also wanted to add to staged-recipes a comment about whether new members wanted to be maintainers. These should be lints, but not something that would fail the build.

jakirkham:
Right. So we could find the commit and comment on it. Admittedly that makes it less visible than commenting in a PR. Though determining the latter is likely error prone. What do you think?

That is a cool idea. I like that too. Though would want to make sure we don't repeatedly ask permission of new maintainers. Any thoughts on how we might solve this?

isuruf:
> Right. So we could find the commit and comment on it. Admittedly that makes it less visible than commenting in a PR. Though determining the latter is likely error prone. What do you think?

That's better. One more thing, we'll have to duplicate the functionality of conda-smithy here, for these stuff

isuruf:
@jakirkham, btw, shall we merge this PR, so that I can try this out a bit?

jakirkham:
Sure. Let's save the other stuff in issues somewhere so that we can come back to them.

jakirkham:
Should add that if this works. We should strip the code from `staged-recipes` that is doing this.

conda_forge_webservices/update_teams.py
```diff
@@ -0,0 +1,27 @@
...
+
+    gh = github.Github(os.environ['GH_TOKEN'])
+    org = gh.get_organization(org_name)
+    remote_repo = org.get_repo(repo_name)
+
+    with tmp_directory() as tmp_dir:
+        repo = Repo.clone_from(remote_repo.clone_url, tmp_dir)
+        subprocess.check_output(["conda", "smithy", "register-github", "--organization", org_name, "--add-teams", tmp_dir])
```
jakirkham: Is it possible that `conda-smithy` doesn't get the `GH_TOKEN` given the way we are calling it?

jakirkham: Fixed in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/99 ).

jakirkham:
So I guess the only thing left is to add webhooks to the rest of the feedstocks.

conda_forge_webservices/utils.py
jakirkham: Unnecessary as this is now merged.

isuruf:
> So I guess the only thing left is to add webhooks to the rest of the feedstocks.

Isn't it better to make this an organizational webhook and filter out non-feedstocks?

jakirkham:
It's not that difficult to add them. Just really need some time to do so.

jakirkham:
Partially addresses issue ( https://github.com/conda-forge/conda-forge.github.io/issues/214 ).

PR #99, opened 2017-09-13
Set github.token for conda-smithy

isuruf:


jakirkham:
Was just thinking about this too. 😄 Feel free to merge when you are done.

buildpack-run.sh
```diff
@@ -17,6 +17,9 @@ bash miniconda.sh -b -p $HOME/.conda
 $HOME/.conda/bin/conda update conda --yes
 $HOME/.conda/bin/conda install -c conda-forge --yes conda-smithy python=3.5 tornado pygithub statuspage
 
+mkdir -p $HOME/.conda-smithy
+echo $GH_TOKEN > $HOME/.conda-smithy/github.token
```
jakirkham: This is how we do it in the [maintenance bot]( https://github.com/conda-forge/conda-forge-maintenance/blob/d26994178dced387290a950af429edb807a4054e/buildpack-run.sh#L85-L87 ). Seems we use `STORAGE_LOCN` instead of `HOME`

jakirkham: If we don't already have it, `STORAGE_LOCN="$(pwd)"`.

jakirkham: Fixed in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/100 ).

PR #101, opened 2017-09-13
Team update tweaks

jakirkham:
Some tweaks to the team update script to address issues raised in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/96 ).

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -118,9 +118,8 @@ class UpdateTeamHookHandler(tornado.web.RequestHandler):
             body = tornado.escape.json_decode(self.request.body)
             repo_name = body['repository']['name']
             owner = body['repository']['owner']['login']
-            ref = body['ref']
-            # Only do anything if we are working with conda-forge, and a push to master.
-            if owner == 'conda-forge' and ref == "refs/heads/master":
+            # Only do anything if we are working with conda-forge.
+            if owner == 'conda-forge':
                 update_teams.update_team(owner, repo_name)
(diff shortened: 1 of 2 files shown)


isuruf:
I wanted to add a feature to remove team members, but this would make it impossible. What do you think? Does removing team members happen at all?

conda_forge_webservices/webapp.py
```diff
@@ -118,9 +118,8 @@ def post(self):
             body = tornado.escape.json_decode(self.request.body)
             repo_name = body['repository']['name']
             owner = body['repository']['owner']['login']
-            ref = body['ref']
-            # Only do anything if we are working with conda-forge, and a push to master.
-            if owner == 'conda-forge' and ref == "refs/heads/master":
+            # Only do anything if we are working with conda-forge.
+            if owner == 'conda-forge':
```
jakirkham: Also we have to checkout the right branch when running the team update step above.

jakirkham:
It does happen, but not super often. For the most part this ends up being easy enough for core to handle manually. Though we mostly were able to do this by collecting a list of who needed to be removed with the cron job on Travis CI, which this does away with.

As a side note, Phil cautioned me against automatically removing people as he had a bad experience where he removed everyone from all teams by accident. This is certainly a concern when automating things.

Personally I'm ok with automating removals. We just have to be careful. We might want to start logging who gets added or removed in the Heroku log.

jakirkham:
Going to close. We can revisit in a new PR if it is still relevant.

PR #115, opened 2017-10-02
Add a webservice for updating feedstocks

jakirkham:
Closes https://github.com/conda-forge/conda-forge-webservices/issues/63

Provides an asynchronous webservice, which listens to push events on the feedstocks. When a push event occurs on a feedstock, the webservice updates the commit that the appropriate submodule points to in the feedstocks repo. This way it is a lot easier to handle updates of various feedstocks throughout the organization on demand instead of running a massive batch script over all of the feedstocks (even when most haven't changed).

Note: This is adapted from the [batch feedstock update script]( https://github.com/conda-forge/conda-forge.github.io/blob/b44a39dbcd919222dee6cae3d7ee82bdf44986ec/scripts/update_feedstocks_submodules.py ).

cc @isuruf

### DIFF
--- a/conda_forge_webservices/update_feedstocks.py
+++ b/conda_forge_webservices/update_feedstocks.py
@@ -1,0 +1,60 @@
+import git
+import github
+import os
+from .utils import tmp_directory
+
+
+def update_feedstock(org_name, repo_name):
+    if not repo_name.endswith("-feedstock"):
+        return
+
+    gh = github.Github(os.environ['GH_TOKEN'])
+
+    gh.get_repo("{}/{}".format(org_name, repo_name))
+    name = repo_name[:-len("-feedstock")]
+
+    with tmp_directory() as tmp_dir:
+        feedstocks_url = (
+            "https://{}@github.com/conda-forge/feedstocks.git"
+            "".format(os.environ["GH_TOKEN"])
+        )
+        feedstocks_repo = git.Repo.clone_from(
+            feedstocks_url,
+            tmp_dir
+        )
+
+        # Get the submodule
+        try:
+            feedstock_submodule = feedstocks_repo.submodule(name)
+        except ValueError:
+            feedstock_submodule = feedstocks_repo.create_submodule(
+                name=name,
+                path=os.path.join("feedstocks", name),
+                url=repo_gh.clone_url,
+                branch="master"
+            )
+
+        # Update the feedstocks submodule
+        feedstock_submodule.update(init=True, recursive=False, force=True)
+        feedstock_submodule.branch.checkout(force=True)
+        feedstock_submodule.update(
+            init=True,
+            recursive=False,
+            force=True,
+            to_latest_revision=True
+        )
+
+        # Submit changes
+        if feedstocks_repo.is_dirty():
+            feedstocks_repo.index.commit("Updated feedstocks submodules. [ci skip]")
+            feedstocks_repo.remote().pull(rebase=True)
+            feedstocks_repo.remote().push()
+
+
+if __name__ == '__main__':
+    import argparse
+    parser = argparse.ArgumentParser()
+    parser.add_argument('org')
+    parser.add_argument('repo')
+    args = parser.parse_args()
+    update_team(args.org, args.repo)
--- a/conda_forge_webservices/update_feedstocks.py
+++ b/conda_forge_webservices/update_feedstocks.py
@@ -24,15 +24,16 @@ def update_feedstock(org_name, repo_name):
         )
 
         # Get the submodule
-        try:
-            feedstock_submodule = feedstocks_repo.submodule(name)
-        except ValueError:
-            feedstock_submodule = feedstocks_repo.create_submodule(
-                name=name,
-                path=os.path.join("feedstocks", name),
-                url=repo_gh.clone_url,
-                branch="master"
-            )
+        feedstock_submodule = feedstocks_repo.create_submodule(
+            name=name,
+            path=os.path.join("feedstocks", name),
+            url=repo_gh.clone_url,
+            branch="master"
+        )
+        # Hack needed if the submodule already exists.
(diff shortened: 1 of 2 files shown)


conda_forge_webservices/update_feedstocks.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def update_feedstock(org_name, repo_name):
+    if not repo_name.endswith("-feedstock"):
+        return
+
+    gh = github.Github(os.environ['GH_TOKEN'])
+    org = gh.get_organization(org_name)
+
+    repo_gh = org.get_repo(repo_name)
```
isuruf: You can merge the 2 above lines to `gh.get_repo("{}/{}".format(org_name, repo_name))`

conda_forge_webservices/update_feedstocks.py
```diff
@@ -0,0 +1,63 @@
...
+    name = repo_name.rstrip("-feedstock")
```
isuruf: This doesn't do what you want.
```
In [20]: r = "symengine-feedstock"

In [21]: r.rstrip("-feedstock")
Out[21]: 'symengin'
```

jakirkham: Good catch! Thanks. Have changed it to `r[:-len("-feedstock")]`, which has the right effect for that case.

conda_forge_webservices/update_feedstocks.py
```diff
@@ -0,0 +1,62 @@
...
+            feedstocks_url,
+            tmp_dir
+        )
+
+        # Get the submodule
+        try:
+            feedstock_submodule = feedstocks_repo.submodule(name)
+        except ValueError:
+            feedstock_submodule = feedstocks_repo.create_submodule(
+                name=name,
```
jakirkham: In theory, we should be able to use this even when the feedstock already exists. However it appears that when the feedstock already exists, we get an issue about the `name` not being defined. 😕 Have raised upstream as issue ( https://github.com/gitpython-developers/GitPython/issues/678 ).

jakirkham: Seems that one can hack around this issue by doing `feedstock_submodule._name = name`.

jakirkham: Used the workaround sent upstream in PR ( https://github.com/gitpython-developers/GitPython/pull/679 ). We can drop it after the next `gitpython` release.

jakirkham:
Please let me know if there is anything else @isuruf.

jakirkham:
Thanks @isuruf. Good point. Have just done so now.

jakirkham:
Looks like it is passing. 😄

jakirkham:
This is good timing to as the feedstock update script has been down for the past couple of days. Likely due to the sheer number of feedstocks it is trying to update.

conda_forge_webservices/update_feedstocks.py
```diff
@@ -0,0 +1,61 @@
...
+from .utils import tmp_directory
+
+
+def update_feedstock(org_name, repo_name):
+    if not repo_name.endswith("-feedstock"):
+        return
+
+    gh = github.Github(os.environ['GH_TOKEN'])
+
+    gh.get_repo("{}/{}".format(org_name, repo_name))
```
jakirkham: Somehow missed storing this anywhere. 😱 Fixing in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/117 ).

conda_forge_webservices/update_feedstocks.py
```diff
@@ -0,0 +1,61 @@
...
+        feedstock_submodule.update(
+            init=True,
+            recursive=False,
+            force=True,
+            to_latest_revision=True
+        )
+
+        # Submit changes
+        if feedstocks_repo.is_dirty():
+            feedstocks_repo.index.commit("Updated feedstocks submodules. [ci skip]")
```
jakirkham: Was hoping the GitHub token would be enough to get the author info set correctly, but appears that is not the case. Adding author details in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/118 ).

jakirkham: Need to manually stage changes as well for `commit` to work. Addressed in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/119 ).

jakirkham:
Took a few follow-up PRs, but it appears it is now working. 🎉

Might be worth adding a command work to update the feedstocks repo if it is out of date.

jakirkham:
Partially addresses issue ( https://github.com/conda-forge/conda-forge.github.io/issues/214 ).

conda_forge_webservices/update_feedstocks.py
isuruf: Would be nice if the commit message had the repo name.

jakirkham: Was thinking about that also. Didn't want to deviate from the feedstock update script too much (at least not in behavior) initially. Though once we get rid of that script, seems reasonable to update this to be more flexible.

jakirkham: PR ( https://github.com/conda-forge/conda-forge-webservices/pull/134 ) should do the trick.

conda_forge_webservices/update_feedstocks.py
```diff
@@ -0,0 +1,61 @@
...
+            feedstocks_repo.remote().push()
+
+
+if __name__ == '__main__':
+    import argparse
+    parser = argparse.ArgumentParser()
+    parser.add_argument('org')
+    parser.add_argument('repo')
+    args = parser.parse_args()
+    update_team(args.org, args.repo)
```
jakirkham: Should have been `update_feedstock`. Fixing in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/135 ).

Issue #121, opened 2017-10-02
Update feedstock listing with feedstocks step via staged-recipes


jakirkham:
Would be neat if we could trigger a webhook from staged-recipes after conversion to update the feedstock listing using the feedstock service. Ideally we could actually contain this listing in the feedstocks repo. This way we can update it completely independently of the docs page. Also the listing would be available near the feedstocks it is connected to. Given how GitHub Pages are setup, it would still be possible to view the feedstocks listing at the same URL we already link to.

jakirkham:
The old way that we handled this was running [`.CI/trigger_travis_conda-forge.github.io.py`]( https://github.com/conda-forge/staged-recipes/blob/24fca413071590dfce3926e4a0cd64dbd7262140/.CI/trigger_travis_conda-forge.github.io.py ) once [conversion completed]( https://github.com/conda-forge/staged-recipes/blob/24fca413071590dfce3926e4a0cd64dbd7262140/.CI/create_feedstocks#L15 ). We can certainly revisit the same strategy.

That said, the point is we only need to regenerate the feedstock listing when there are new feedstocks being added. So it is a bit silly to run that script every time we merge something at the webpage repo. One way to fix that would be to run the script through a webservice instead that only staged-recipes triggers.

Another point is that we don't really need that listing in the webpage repo at all as it only pertains to the feedstocks. So it might make more sense to include the listing in the feedstocks repo for a better separation of concerns.

jakirkham:
Proposing to bring back the old way of handling this issue in PR ( https://github.com/conda-forge/staged-recipes/pull/4107 ) until we have a new solution.

jakirkham:
Should be addressed with PR ( https://github.com/conda-forge/conda-forge-webservices/pull/142 ).

jakirkham:
Dropping the webpage listing and its generation during CI in PR ( https://github.com/conda-forge/conda-forge.github.io/pull/466 ). Also dropping the webpage trigger at staged-recipes with PR ( https://github.com/conda-forge/staged-recipes/pull/4184 ).

PR #127, opened 2017-10-03
Print rate limiting info

isuruf:
Copied code from `staged-recipes`

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -14,6 +14,7 @@ import github
 import conda_smithy.lint_recipe
 import shutil
 from contextlib import contextmanager
+from datetime import datetime
 
 
 import conda_forge_webservices.linting as linting
@@ -23,7 +24,7 @@ import conda_forge_webservices.update_teams as update_teams
 import conda_forge_webservices.commands as commands
 
 
-def print_rate_limiting_info():
+def print_rate_limiting_info_for_token(token, user):
     # Compute some info about our GitHub API Rate Limit.
     # Note that it doesn't count against our limit to
     # get this info. So, we should be doing this regularly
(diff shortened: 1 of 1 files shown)


jakirkham:
Maybe we should make the token an argument. Would be useful if we start using multiple.

isuruf:
Then I'd guess we should print info for everything.

isuruf:
Fixed it and pushed to a new branch in this repo

conda_forge_webservices/webapp.py
```diff
@@ -51,7 +51,6 @@ def print_rate_limiting_info():
 
     d = [
          (os.environ['GH_TOKEN'], "conda-forge-linter"),
-         (os.environ['FEEDSTOCKS_GH_TOKEN'], "conda-forge-coordinator"),
```
jakirkham: We still need to push though. 😕

isuruf: Git operations don't count towards GitHub API

jakirkham: Good to know. Must have missed the documentation for that. Though we may start using the API to handle issue ( https://github.com/conda-forge/conda-forge-webservices/issues/121 ).

conda_forge_webservices/webapp.py
```diff
@@ -23,6 +24,42 @@
...
+    gh_api_reset_time = gh.get_rate_limit().rate.reset
+    gh_api_reset_time -= datetime.utcnow()
+    msg = "{user} - remaining {remaining} out of {total}.".format(remaining=gh_api_remaining,
+            total=gh_api_total, user=user)
+    print("-"*len(msg))
+    print(msg)
+    print("Will reset in {time}.".format(time=gh_api_reset_time))
+
+
+def print_rate_limiting_info():
```
jakirkham: Maybe we can just make token and name arguments here too. That way we only need print the one relevant for the different services. This should play more nicely as we add more tokens.

isuruf: This way it's easier to figure out from the heroku logs. Besides the call to check rate limit is not counted towards the limit.

jakirkham: Maybe we should look into logging instead of using `print`, but that is a separate discussion.

jakirkham:
Thanks @isuruf.

Issue #122, opened 2017-10-02
Different GitHub tokens for each service


jakirkham:
To avoid bogging down the different services gathered here, we should try to get them to use different GitHub tokens. Right now they all use the linter's token. However we can at least get team updates and feedstock updates to use the coordinator's token as was done in the webpage repo. We may need to check what other bots are used for and add them as needed. This all in an attempt to avoid going over GitHub's rate limit.

isuruf:
+1. Note that the webservices using `conda_smithy` will have to use the same token.

isuruf:
Maybe we should print the rate limit in heroku log after each webservice to get an idea of how big of an issue this is going to be.

jakirkham:
That's a good idea. There's a few lines of code at staged-recipes that could be pilfered for this purpose.

jakirkham:
As for the `conda-smithy` point, maybe we can swap out the tokens somehow while services are running. Alternatively we can investigate some override mechanism for `conda-smithy`.

jakirkham:
Fixed feedstock updates to use a different token. Though it is shared by the webpage repo.

beckermr:
The code now uses an app token when possible which has a rate limit that is >2x a normal token. Closing this as addressed.

PR #129, opened 2017-10-03
Use a different token for feedstock updates

jakirkham:
Partially addresses issue ( https://github.com/conda-forge/conda-forge-webservices/issues/122 ).

Instead of using the token that is associated with the linter use a token associated with the coordinator. This matches more closely to the behavior that was present when feedstock updates were performed by the webpage repo's CI. Should help mitigate rate limiting for the linter and other services due to feedstock updates.

cc @isuruf


jakirkham:
Missed a token here. Fixing in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/130 ).

PR #134, opened 2017-10-03
Named feedstock updates

jakirkham:
When updating the feedstocks repo, includes the name of the feedstock updated in each commit message.

cc @isuruf

### DIFF
--- a/conda_forge_webservices/update_feedstocks.py
+++ b/conda_forge_webservices/update_feedstocks.py
@@ -52,7 +52,7 @@ def update_feedstock(org_name, repo_name):
             )
             feedstocks_repo.git.add(update=True)
             feedstocks_repo.index.commit(
-                "Updated feedstocks submodules.",
+                "Updated the {0} feedstock.".format(name),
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/update_feedstocks.py
```diff
@@ -52,7 +52,7 @@ def update_feedstock(org_name, repo_name):
             )
             feedstocks_repo.git.add(update=True)
             feedstocks_repo.index.commit(
-                "Updated feedstocks submodules. [ci skip]",
+                "Updated the %s feedstock." % name,
```
isuruf: I prefer to use one style of formatting `.format` or `%` in a project.

jakirkham: Yeah was thinking about that too. Bad habit on my part.

Also was thinking that it would be nice to distinguish a new addition from an updated one.

jakirkham: Switched to `.format`.

PR #136, opened 2017-10-03
Avoid calling GitHub API when updating feedstocks

jakirkham:
Based on the URLs of the other feedstocks in `.gitmodules` both before the webservice and after, they all seem to have the same format. It is pretty easy to create a correctly formatted URL for updating these feedstock modules without using the GitHub API. Hence we drop the unneeded GitHub API calls and format the feedstock URL ourselves.

### DIFF
--- a/conda_forge_webservices/update_feedstocks.py
+++ b/conda_forge_webservices/update_feedstocks.py
@@ -1,5 +1,4 @@
 import git
-import github
 import os
 from .utils import tmp_directory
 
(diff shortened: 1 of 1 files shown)


isuruf:
I think there are many places where we could avoid github API. +1 for this effort.

jakirkham:
Thanks for taking a look.

Yeah it was my bad that I was using it at all here. Think it was just the result of something that was (nearly) optimized away.

jakirkham:
FWIW seems this makes it run a bit faster too. 😄

PR #137, opened 2017-10-03
Linting force comment

isuruf:


jakirkham:
Should it be forced? Am thinking of cases where `[ci skip]`/`[ci skip]` or `[lint skip]`/`[skip lint]` are used.

isuruf:
This is for the case of command `please lint`

jakirkham:
Right. However I wouldn't expect to run the linter if the commit messages explicitly disable it. Just as I wouldn't expect CI to be run on a PR where I have explicitly skipped them via commit messages.

isuruf:
I was thinking of a case like, https://github.com/conda-forge/conda-forge-build-setup-feedstock/pull/69#issuecomment-333913863 where there is no feedback

jakirkham:
It looks like it is just reusing the comment from before. That said, have no problem with adding an explicit comment from the linter when requested. Makes perfect sense to me.

Just would also expect the linter to respect the skip notes in commit messages as well. Perhaps it could say lint was skipped by commit message if we want a response for that case too.

jakirkham:
...or is this already caught by the [`compute_lint_message`]( https://github.com/conda-forge/conda-forge-webservices/pull/137/files#diff-003f6ff938ed739a0d10990176da3f78R139 )?

isuruf:
> ...or is this already caught by the compute_lint_message?

Yes, it is

jakirkham:
LGTM. Good to merge?

jakirkham:
Just realized we didn't have CI. Pushed to the repo to confirm it passes. Not expecting any surprises though. Feel free to merge when it passes.

jakirkham:
Thanks @isuruf.

PR #142, opened 2017-10-06
Refresh feedstock listing

jakirkham:
Closes https://github.com/conda-forge/conda-forge-webservices/issues/121

Refresh the webpage's feedstock listing within the `feedstocks` repo. This should help cutdown the noise on the webpage repo a bit. Also should better collect related items. Namely the feedstock listing will live with the feedstocks and should be more performant as a result.

The displayed result should be the same once we enable the `feedstocks` repo to have GitHub Pages working on its `master` branch. Links from the main webpage have already been updated to support this behavior. Style sheets, icons, and javascript code should still be pulled from the main webpage. Though this last point is difficult to test without making this live. The template for the listing itself has been copied from the webpage repo.

Initially this borrowed code from `conda-smithy` and used the GitHub API. However it has been optimized to simply list the contents of the `feedstocks` repo and use this to generate the HTML page. Based on local testing, this appears to be quite fast. Thus we avoid running into rate limiting with this approach unlike the current approach. Not to mention this does not require CI time to build.

Updates will be triggered by pushes to `staged-recipes`. Admittedly not all pushes to `staged-recipes` will trigger an actionable change. Only when feedstocks are created. There is likely some room for optimization there, but will require some clever parsing of the commit message or git history at `staged-recipes` to achieve. For now optimization on this front is skipped. Considering the other optimizations this already provides (freeing up CIs, speeding up generation), this seems reasonable.


jakirkham:
Any thoughts on this @isuruf? 😉

isuruf:
I like that this is making incremental changes. What I don't is generating the html here. How about running a cron job on the webpage repo that updates the generated html using the feedstocks repo?

jakirkham:
What is it that you don't like about having it here?

isuruf:
My main issue is that feedstocks.html.tmpl is checked into this repo. That file doesn't belong in this repo

jakirkham:
Ok, I think I understand that concern. Is this basically a concern that it will get out of sync with the rest of the webpage repo content? If so, what if we download that file here from the webpage repo?

isuruf:
Yes, that would work

jakirkham:
Thanks @isuruf. I think these changes will be more to your liking then.

We will still need to update the template back at the webpage repo to get these changes to work correctly. Though this can go in independently of that change. After all this won't go live until we add the webhook to staged-recipes.

jakirkham:
Tweaked it to work with the existing `feedstocks.html.tmpl` to keep things simple.

jakirkham:
How does this look to you @isuruf?

isuruf:
Looks good to me.

jakirkham:
Thanks @isuruf. Going to take this for a spin.

jakirkham:
Forgot about adding a `.nojekyll` file. Have done that in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/145 ).

conda_forge_webservices/feedstocks_service.py
```diff
@@ -0,0 +1,126 @@
...
+import jinja2
+import os
+
+from collections import namedtuple
+from .utils import tmp_directory
+
+
+def handle_feedstock_event(org_name, repo_name):
+    if repo_name == "staged-recipes":
+        update_listing()
```
jakirkham: Since we are using the template from the webpage repo, we should also pick up on pushes to the webpage repo. That way if the feedstocks listing template changes, we can refresh the feedstocks listing accordingly. Making this change in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/146 ).

jakirkham:
Ended up needing to publish the listing to `feedstocks` repo's `gh-pages` branch instead of `master`. Was running into issues where GitHub pages timed out trying to clone the feedstocks repo's master branch or had issues cloning a submodule for some reason. Switched to `gh-pages` in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/147 ), which has fixed this issue. The feedstocks listing is now live again.

PR #146, opened 2017-10-15
Update feedstocks listing on webpage update

jakirkham:
If the webpage repo is updated, also update the feedstocks listing. This is done because the feedstocks listing template could be updated in the webpage repo, which could result in changes to the feedstocks listing. Hence we should try regenerating the feedstocks listing in this case so as to make any needed changes to the format.


jakirkham:
Already have added the webhook to the webpage repo for this service. So support for this change should go live as soon as this is merged.

PR #147, opened 2017-10-15
Place `feedstocks` listing under `gh-pages`

jakirkham:
Instead of using the `feedstocks` repo's `master` branch for publishing the feedstocks listing webpage, use the `gh-pages` branch. As it seems GitHub tries to clone all submodules when generating GitHub Pages for the `feedstocks` repo, this fails pretty easily and often. It would be better to use a different branch in the `feedstocks` repo to avoid having to deal with any submodule related issues. Hence this is what we do here.


jakirkham:
Cleaning out the old webpage listing content from the `feedstocks` repo's `master` branch in PR ( https://github.com/conda-forge/feedstocks/pull/18 ).

PR #148, opened 2017-10-22
Use `git add -A` to update feedstocks

jakirkham:
Was previously using `git add -u`. However that might miss some things that need to be added, which `git add -A` should catch (e.g. previously untracked changes).


jakirkham:
Part of the motivation here is I am seeing empty commits like commit ( https://github.com/conda-forge/feedstocks/commit/bebc319a02189803488f7af61344992db38abe72 ) pushed to the `feedstocks` repo, which indicates the previous `git add -u` was causing some changes to be unstaged. That or there are some formatting changes that need to happen.

jakirkham:
Doesn't seem to be the right fix. So closing this out.

jakirkham:
Replaced with PR ( https://github.com/conda-forge/conda-forge-webservices/pull/149 ).

PR #149, opened 2017-10-22
Stage changes and commit if the index is dirty

jakirkham:
Make sure to stage changes first and only create a commit if there are changes staged in the index (not if there are ones in the working tree). It is possible to have changes that can't easily be staged if there are formatting issues (e.g. `.gitattributes` related changes) that need to be addressed in a feedstock for instance. This provides some extra assurance against creating empty commits.

### DIFF
--- a/conda_forge_webservices/feedstocks_service.py
+++ b/conda_forge_webservices/feedstocks_service.py
@@ -65,16 +65,17 @@ def update_listing():
         with open(nojekyll, 'w') as fh:
             pass
 
-        if feedstocks_page_repo.is_dirty(untracked_files=True):
+        feedstocks_page_repo.index.add([os.path.relpath(
+            feedstocks_html, feedstocks_page_dir
+        )])
+        feedstocks_page_repo.index.add([os.path.relpath(
+            nojekyll, feedstocks_page_dir
+        )])
+
+        if feedstocks_page_repo.is_dirty(working_tree=False, untracked_files=True):
             author = git.Actor(
                 "conda-forge-coordinator", "conda.forge.coordinator@gmail.com"
             )
-            feedstocks_page_repo.index.add([os.path.relpath(
-                feedstocks_html, feedstocks_page_dir
-            )])
-            feedstocks_page_repo.index.add([os.path.relpath(
-                nojekyll, feedstocks_page_dir
-            )])
             feedstocks_page_repo.index.commit(
                 "Updated the feedstock listing.",
                 author=author,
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/feedstocks_service.py
```diff
@@ -118,13 +119,13 @@ def update_feedstock(org_name, repo_name):
             force=True,
             to_latest_revision=True
         )
+        feedstocks_repo.git.add([".gitmodules", feedstock_submodule.path])
```
jakirkham: Figured we should be more strict about what we are adding as well. In general, these are the only two things that should be changing. The first only changes if we are adding a new feedstock. The second always changes if there is something to commit.

conda_forge_webservices/feedstocks_service.py
```diff
@@ -118,13 +119,13 @@ def update_feedstock(org_name, repo_name):
...
 
         # Submit changes
-        if feedstocks_repo.is_dirty(untracked_files=True):
+        if feedstocks_page_repo.is_dirty(working_tree=False, untracked_files=True):
```
jakirkham: This means we only check whether there are changes in the index. So if `git add` above included anything, we will proceed to commit. If not, there is nothing to commit (as nothing was staged).

conda_forge_webservices/feedstocks_service.py
```diff
@@ -65,16 +65,17 @@ def update_listing():
...
 
-        if feedstocks_page_repo.is_dirty(untracked_files=True):
+        feedstocks_page_repo.index.add([os.path.relpath(
+            feedstocks_html, feedstocks_page_dir
+        )])
+        feedstocks_page_repo.index.add([os.path.relpath(
+            nojekyll, feedstocks_page_dir
+        )])
+
+        if feedstocks_page_repo.is_dirty(working_tree=False, untracked_files=True):
```
jakirkham: Can't imagine why there would be any changes that are unstaged here, but thought this check is better than allowing empty commits through.

jakirkham:
Going to merge this to test it while I still have some time to do so. Though still welcome feedback should anything need to be changed.

conda_forge_webservices/feedstocks_service.py
jakirkham: The `feedstocks_repo` here should not have changed to `feedstocks_page_repo`. Fixing in PR ( https://github.com/conda-forge/conda-forge-webservices/pull/150 ).

jakirkham:
With the exception of the minor typo noted above (since fixed), this worked as intended. Namely the `dj-static` repo did not create another empty commit in the `feedstocks` repo due to its unstaged changes.

Issue #151, opened 2017-11-01
Runaway team update bot


jakirkham:
Looks like the bot got a little overexcited about updating the team in issue ( https://github.com/conda-forge/pandoc-feedstock/issues/21 ). Not sure what happened, but am pretty sure it is unintended.

pkgw:
See also conda-forge/healpy-feedstock#9.

pkgw:
Ahhhh I made it start comment again!!! No!!! Sorry @mwcraig.

pkgw:
Based on what just happened, maybe it will make a comment whenever it sees any comment on the relevant issue, rather than just when it sees a comment requesting a team update?

isuruf:
Thanks, I was confused why this happened, but the report on what happened today was helpful in figuring it out

PR #160, opened 2018-01-25
Fix loop

isuruf:
When a new comment is made, body['issue']['body'] gives the first comment of the issue
which makes the bot confused.

Fixes https://github.com/conda-forge/conda-forge-webservices/issues/151. cc @jakirkham, @pkgw

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -244,7 +244,10 @@ class CommandHookHandler(tornado.web.RequestHandler):
 
             if not pull_request and action in ['opened', 'edited', 'created', 'reopened']:
                 title = body['issue']['title'] if event == "issues" else ""
-                comment = body['issue']['body']
+                if 'comment' in body:
+                    comment = body['comment']['body']
+                else:
+                    comment = body['issue']['body']
                 commands.issue_comment(owner, repo_name, issue_num, title, comment)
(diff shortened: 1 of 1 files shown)


pkgw:
I think there will also be a loop if the magic "please ..." command is in the issue title. It checks for stuff to do when an issue is edited in any way ([ref](https://github.com/conda-forge/conda-forge-webservices/blob/master/conda_forge_webservices/webapp.py#L228)) and part of that checking is to see whether the issue title contains a command ([ref](https://github.com/conda-forge/conda-forge-webservices/blob/master/conda_forge_webservices/commands.py#L75)). If the title doesn't change, I think it will keep on doing what it was told.

isuruf:
'title' is set to empty when it is an issue_comment, so there will be no loop there

conda_forge_webservices/webapp.py
```diff
@@ -244,7 +244,10 @@ def post(self):
 
             if not pull_request and action in ['opened', 'edited', 'created', 'reopened']:
                 title = body['issue']['title'] if event == "issues" else ""
```
isuruf: cc @pkgw

isuruf:
merging this to test it out

isuruf:
Looks like it worked. @pkgw, I changed the webhook to stop the torrent of comments, and restored it now to check

PR #166, opened 2018-03-07
Add second positional arg to rerender call

### LINKED ISSUE #165: Missing argument in call to `rerender()`
Seems like we're missing a positional arg for ``pr_num`` [here in commands.py](https://github.com/conda-forge/conda-forge-webservices/blob/master/conda_forge_webservices/commands.py#L119).  I'm not 100% sure what this should be, exactly. It seems like rerender is expecting an active pull request when it is called but the call to rerender linked above appears to be working on local repo before the PR has been created on GH.

ericdill:
Closes #165

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -116,7 +116,7 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
 
         if ADD_NOARCH_MSG in comment + title:
             make_noarch(git_repo)
-            rerender(git_repo)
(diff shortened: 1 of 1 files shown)


jakirkham:
Pushed to the repo (as that's how we test it).

ericdill:
@jakirkham oh are branches supposed to be pushed to the conda-forge repo and not my fork?

jakirkham:
The branches on the repo are tested using a token.

Normally I do both. But yeah if we don't push to the repo, the CI is less informative.

Issue #43, opened 2016-08-24
Linter failing


pelson:
Is this a transient issue?

```
HTTPServerRequest(protocol='http', host='conda-forge.herokuapp.com', method='POST', uri='/conda-linting/hook', version='HTTP/1.1', headers={'X-Github-Event': 'pull_request', 'Total-Route-Time': '0', 'Accept': '*/*', : 'http', ...})
2016-08-24T08:16:24.761335+00:00 app[web.1]: Traceback (most recent call last):
2016-08-24T08:16:24.761340+00:00 app[web.1]:     issue = repo.get_issue(pr_id)
2016-08-24T08:16:24.761339+00:00 app[web.1]:   File "/app/conda_forge_webservices/linting.py", line 27, in compute_lint_message
2016-08-24T08:16:24.761338+00:00 app[web.1]:   File "/app/conda_forge_webservices/webapp.py", line 78, in post
2016-08-24T08:16:24.761340+00:00 app[web.1]:   File "/app/.conda/lib/python3.5/site-packages/github/Repository.py", line 1599, in get_issue
2016-08-24T08:16:24.761340+00:00 app[web.1]:     self.url + "/issues/" + str(number)
2016-08-24T08:16:24.761339+00:00 app[web.1]:     lint_info = linting.compute_lint_message(owner, repo_name, pr_id)
2016-08-24T08:16:24.761341+00:00 app[web.1]:   File "/app/.conda/lib/python3.5/site-packages/github/Requester.py", line 172, in requestJsonAndCheck
2016-08-24T08:16:24.761792+00:00 app[web.1]: ERROR:tornado.access:500 POST /conda-linting/hook (192.30.252.45) 184.69ms
2016-08-24T08:16:24.761343+00:00 app[web.1]: github.GithubException.UnknownObjectException: 404 {'documentation_url': 'https://developer.github.com/v3', 'message': 'Not Found'}
2016-08-24T08:16:24.761342+00:00 app[web.1]:     return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
2016-08-24T08:16:24.761342+00:00 app[web.1]:   File "/app/.conda/lib/python3.5/site-packages/github/Requester.py", line 180, in __check
2016-08-24T08:16:24.761342+00:00 app[web.1]:     raise self.__createException(status, responseHeaders, output)
```

Is this the reason for https://github.com/conda-forge/bob.io.image-feedstock/pull/17#issuecomment-241990912?

[...]

Issue #30, opened 2016-06-26
Random failure of the service


pelson:
For example, the PR https://github.com/conda-forge/conda-build-all-feedstock/pull/11 failed to lint with the following traceback (and resulting 500 http code):

```
2016-06-26T08:50:13.693245+00:00 heroku[router]: at=info method=POST path="/conda-linting/hook" host=conda-forge.herokuapp.com request_id=5c7283...5d2b fwd="192.30.252.46" dyno=web.1 connect=0ms service=200ms status=500 bytes=255
2016-06-26T08:50:13.697342+00:00 app[web.1]: ERROR:tornado.application:Uncaught exception POST /conda-linting/hook (192.30.252.46)
2016-06-26T08:50:13.697355+00:00 app[web.1]: HTTPServerRequest(protocol='http', host='conda-forge.herokuapp.com', method='POST', uri='/conda-linting/hook', version='HTTP/1.1', remote_ip='192....46', headers={'Host': 'conda-forge.herokuapp.com', 'X-Request-Start': '1466....3491', 'Content-Length': '23052', 'Via': '1.1 vegur', 'X-Github-Event': 'pull_request', 'X-Github-Delivery': 'fee75...961881de', 'X-Forwarded-Port': '80', 'X-Forwarded-For': '192.....46', 'X-Request-Id': '5c7283...ed0f37e95d2b', 'Accept': '*/*', 'User-Agent': 'GitHub-Hookshot/b63f7ac', 'Total-Route-Time': '0', 'Connection': 'close', 'X-Forwarded-Proto': 'http', 'Connect-Time': '0', 'Content-Type': 'application/json'})
2016-06-26T08:50:13.697357+00:00 app[web.1]: Traceback (most recent call last):
2016-06-26T08:50:13.697358+00:00 app[web.1]:   File "/app/.conda/lib/python2.7/site-packages/tornado/web.py", line 1443, in _execute
2016-06-26T08:50:13.697359+00:00 app[web.1]:     result = method(*self.path_args, **self.path_kwargs)
2016-06-26T08:50:13.697361+00:00 app[web.1]:   File "/app/conda_forge_webservices/webapp.py", line 61, in post
2016-06-26T08:50:13.697361+00:00 app[web.1]:     lint_info = linting.compute_lint_message(owner, repo_name, pr_id)
2016-06-26T08:50:13.697362+00:00 app[web.1]:   File "conda_forge_webservices/linting.py", line 27, in compute_lint_message
[...]

PR #169, opened 2018-03-19
Wait up to 1 second before linting

jakirkham:
Closes https://github.com/conda-forge/conda-forge-webservices/issues/30
Closes https://github.com/conda-forge/conda-forge-webservices/issues/43
Closes https://github.com/conda-forge/conda-forge-webservices/issues/126

Appears the webhooks for PRs fire a bit in advance of when the merge ref shows up. While this was solved by sleeping for 0.1 seconds at a time, it appears that is no longer enough. This sleeps for 1 second at a time in the hope that will be enough time for the merge ref to show up.

cc @isuruf

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -22,7 +22,7 @@ def compute_lint_message(repo_owner, repo_name, pr_id, ignore_base=False):
 
     mergeable = None
     while mergeable is None:
-        time.sleep(0.1)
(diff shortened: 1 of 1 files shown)


pelson:
Thanks for tweaking this a bit @jakirkham. Reminds me that I've long intended to go back an async all the things here - I imagine this change has the potential to increase the number of request timeouts, so worth watching out for it.

Thanks again,

Issue #125, opened 2017-10-03
Increment build number in "add noarch: python"


isuruf:
cc @sodre, @jakirkham

CJ-Wright:
Also would it be possible to set the install command to use pip?

jakirkham:
Slight preference for adding `noarch: python` after `build` as well just to keep the build number at the top of the section. That said, know this stuff is a pain. So can be reasonable. :)

Issue #112, opened 2017-09-25
Command to rerender in issue


isuruf:
Currently command to rerender is limited to PRs. Should we extend this to issues as well? By adding this to issues, we are making it very easy to rerender and this might result in more CI traffic for no good reason.

cc @ocefpaf

ocefpaf:
> Currently command to rerender is limited to PRs. Should we extend this to issues as well? By adding this to issues, we are making it very easy to rerender and this might result in more CI traffic for no good reason.

I guess it is CI limitation vs being human friendly. Let's err on the side of stability and not do that right now. Do you agree?

isuruf:
I'm not sure anymore. There are cases where we need to rerender for python and r version upgrades.
FYI. this is only rerender. I've enabled `noarch: python`, because it saves CI.

ocefpaf:
If you want to implement that and see how it'll behave I am fine with that. We can always pull the service later.

pkgw:
[The docs](https://conda-forge.org/docs/webservice.html#conda-forge-admin-please-rerender) imply that this feature is available. If so it behaves a bit unintuitively: you can't trigger a rerender when *creating* an issue, but only when commenting on a preexisting one: [example](https://github.com/conda-forge/python-casacore-feedstock/issues/6).

pkgw:
Also, in the example linked above, rerendering locally *does* in fact cause changes to the recipe.

jakirkham:
So this should work in a new issue. Here's an [example]( https://github.com/conda-forge/vigra-feedstock/issues/41 ). If it didn't work though, there may be a bug or a transient issue. Would suggest trying again. If it persists, let's move it to a new issue here.

Tracking the issues down is kind of a chore as we can't really share the logs publicly for security reasons. It would be nice to fix that. In any event, if the issue is repeatable, a core member can step and debug this pretty easily.

jakirkham:
Closing as this is completed.

PR #170, opened 2018-03-19
make bot commands more flexible / search more precise

dougalsutherland:
I never remember whether the command is "please rerender" or "please re-render"; this change makes both work. It also makes "please add `noarch: python`", "please make noarch python", etc valid too in addition to just "please add noarch: python".

While I was making that change, I realized that the current check was a little unintuitive. For example, if I tagged conda-forge-admin here, then it would actually add noarch (if this were a feedstock) because I said the command above (completely separated from the tag). That's not incredibly likely to happen, but still, this change stops that: you now have to tag the bot, optionally add a comma/colon/whitespace, then immediately say the command.

(I didn't add tests for these, since there weren't any for the commands in the first place.)

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -67,6 +67,7 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
     text = comment + title
 
     issue_commands = [UPDATE_TEAM_MSG, ADD_NOARCH_MSG, UPDATE_CIRCLECI_KEY_MSG]
+    send_pr_commands = [ADD_NOARCH_MSG]
 
     if not any(command.search(text) for command in issue_commands):
         return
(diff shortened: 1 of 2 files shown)


jakirkham:
Seems fine in principle. Any thoughts on tests for these changes?

dougalsutherland:
I can try adding some tests with comments that should / shouldn't trigger commands, with the actual edits mocked out, later on.

jakirkham:
Sure. Yeah even a check to see that it would trigger the bot would be good.

Depending on what you have in mind, [these tests]( https://github.com/conda-forge/conda-forge-webservices/blob/master/conda_forge_webservices/tests/linting/test_compute_lint_message.py ) might be useful to steal from.

dougalsutherland:
Added some tests, and in doing so caught some minor bugs in the current logic. :)

dougalsutherland:
This is a little beyond the scope of the initial PR now, but I also went ahead and added support for opening a `please rerender` issue, for the occasional situation where the only thing wrong is that the rendering needs to be updated.

jakirkham:
Thanks for looking into these things. Sounds like you are fixing some issues that we have been slow to address thus far, but are aware of. References below.

> This is a little beyond the scope of the initial PR now, but I also went ahead and added support for opening a `please rerender` issue, for the occasional situation where the only thing wrong is that the rendering needs to be updated.

Sounds like issue ( https://github.com/conda-forge/conda-forge-webservices/issues/112 ).

> Both that and `add noarch python` should probably bump the version number, though....

Sounds like issue ( https://github.com/conda-forge/conda-forge-webservices/issues/125 ).

jakirkham:
Could you also push these changes to this repo under the branch name `dougalsutherland/command-flexibility`? Most of the tests require a test token, which is not provided in fork CI builds (understandably).

dougalsutherland:
@conda-forge/conda-forge-webservices, this is ready for review. :)

isuruf:
Thanks @dougalsutherland

jakirkham:
FWIW IMHO we should add the re-rendering commands to issues as there are enough scenarios where I or other did not have a PR, but needed a re-render. Would be helpful in these scenarios to just allow the command and have the bot do the work.

isuruf:
+1

dougalsutherland:
Yeah, it just makes more sense to have it bump the build number too, which requires actually parsing the meta.yaml at least somewhat, so I put it off....

isuruf:
Even without bumping the build number, it'll be useful to fix incorrect renderings by users.

jakirkham:
Agreed. If someone wants to put a PR for this, would be +1 on merging.

PR #175, opened 2018-04-05
allow rerender command in issues

isuruf:
cc @dougalsutherland

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -135,6 +135,10 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
                         Here's a checklist to do before merging.
                         - [ ] Bump the build number if needed.
                         """.format(comment_msg, issue_num))
+
+                if to_close:
+                    pr_message += "\nFixes #{}".format(issue_num)
+
                 pr = repo.create_pull(
                     pr_title, pr_message,
                     "master", "{}:{}".format(forked_user, forked_repo_branch))
@@ -145,10 +149,6 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
                         I just wanted to let you know that I {} in {}/{}#{}.
                         """.format(comment_msg, org_name, repo_name, pr.number))
                 issue.create_comment(message)
-
-            # rerender() will have made a comment if there was nothing to do
-            if to_close:
-                issue.edit(state="closed")
 
 
 def rerender(repo, pr_num):
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -127,8 +127,16 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
 
             if changed_anything:
                 git_repo.git.push("origin", forked_repo_branch)
+                pr_message = textwrap.dedent("""
+                        Hi! This is the friendly automated conda-forge-webservice.
+
+                        I've {} as instructed #{}.
+
+                        Here's a checklist to do before merging.
+                        - [ ] Bump the build number if needed.
+                        """.format(comment_msg, issue_num))
                 pr = repo.create_pull(
-                    pr_title, "As instructed in #{}".format(issue_num),
+                    pr_title, pr_message,
                     "master", "{}:{}".format(forked_user, forked_repo_branch))
 
                 message = textwrap.dedent("""
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -66,8 +66,9 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
 
     text = comment + title
 
-    issue_commands = [UPDATE_TEAM_MSG, ADD_NOARCH_MSG, UPDATE_CIRCLECI_KEY_MSG]
-    send_pr_commands = [ADD_NOARCH_MSG]
+    issue_commands = [UPDATE_TEAM_MSG, ADD_NOARCH_MSG, UPDATE_CIRCLECI_KEY_MSG,
+                      RERENDER_MSG]
+    send_pr_commands = [ADD_NOARCH_MSG, RERENDER_MSG]
 
     if not any(command.search(text) for command in issue_commands):
         return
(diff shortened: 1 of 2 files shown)


dougalsutherland:
LGTM. The one thing is whether it should bump the build number itself; there are situations where it won’t matter, but especially once rerendering changes the linings we’re going to want build numbers bumped....just need to steal some logic from the ticking not to do that.

jakirkham:
It's an option. We typically haven't bumped the build number for just a re-rendering (with maybe the exception of changing macOS images on Travis CI previously). Though it would be easy enough to do if it's desired.

In `conda-build` 3, changing dependencies will trigger generation of a different package hash, which will result in a new build. IDK where we landed on uploading these when the build number hasn't changed (maybe against?), but given this context it might be worth thinking about a bit more. This isn't to say we shouldn't bump the build number as well. More to say re-renderings that trigger dependency changes could already generate a new package without a build number bump should we want that.

conda_forge_webservices/commands.py
```diff
@@ -66,8 +66,9 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
 
     text = comment + title
```
jakirkham: As it hasn't mattered until now, I'm guessing this doesn't matter at all, but it might be worthwhile to add a newline between these two (and maybe flip the order ;).

conda_forge_webservices/commands.py
```diff
@@ -104,8 +105,8 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
 
         with tmp_directory() as tmp_dir:
             feedstock_dir = os.path.join(tmp_dir, repo_name)
-            repo_url = "https://{}@github.com/{}/{}.git".format(os.environ['GH_TOKEN'],
-                forked_user, repo_name)
+            repo_url = "https://{}@github.com/{}/{}.git".format(
+                os.environ['GH_TOKEN'], forked_user, repo_name)
```
jakirkham: We could start running `flake8` or `pylint` on this if we are concerned about style. This can be saved for an issue. Just this change made me think of that.

conda_forge_webservices/commands.py
```diff
@@ -114,34 +115,48 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
             if ADD_NOARCH_MSG.search(text):
                 make_noarch(git_repo)
                 rerender(git_repo, issue_num)
+                pr_title = "MNT: Add noarch: python"
+                comment_msg = "made the recipe `noarch: python`"
+                changed_anything = True  # make_noarch always adds a line
```
jakirkham: What if it is already noarch?

conda_forge_webservices/commands.py
```diff
@@ -114,34 +115,48 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
...
+            if changed_anything:
                 git_repo.git.push("origin", forked_repo_branch)
-                msg = "MNT: Add noarch: python"
-                pr = repo.create_pull(msg, "As instructed in #{}".format(issue_num),
-                        "master", "{}:{}".format(forked_user, forked_repo_branch))
-
-                if ADD_NOARCH_MSG.search(title):
-                    issue.edit(state="closed")
+                pr = repo.create_pull(
+                    pr_title, "As instructed in #{}".format(issue_num),
```
jakirkham: Thoughts on having the PR close the issue on merge? I know we are already closing the issue below. Just thought it was worth considering.

conda_forge_webservices/commands.py
```diff
@@ -114,34 +115,48 @@ def issue_comment(org_name, repo_name, issue_num, title, comment):
...
 
-                        I just wanted to let you know that I made the recipe noarch: python in {}/{}#{}.
-                        """.format(org_name, repo_name, pr.number))
+                        I just wanted to let you know that I {} in {}/{}#{}.
+                        """.format(comment_msg, org_name, repo_name, pr.number))
                 issue.create_comment(message)
 
+            # rerender() will have made a comment if there was nothing to do
+            if to_close:
+                issue.edit(state="closed")
```
jakirkham: Please see comment above.

isuruf:
> In conda-build 3, changing dependencies will trigger generation of a different package hash, which will result in a new build. IDK where we landed on uploading these when the build number hasn't changed (maybe against?), but given this context it might be worth thinking about a bit more. This isn't to say we shouldn't bump the build number as well. More to say re-renderings that trigger dependency changes could already generate a new package without a build number bump should we want that.

Upload script will not upload a package if there is a package with the same version+build_num regardless of the hash

conda_forge_webservices/commands.py
isuruf: https://github.com/dougalsutherland/conda-forge-webservices/commit/7e071b6ce8f5ad05a38d8e415577831a42a0b9dd

conda_forge_webservices/tests/test_commands.py
```diff
@@ -115,6 +115,17 @@ def test_issue_command_triggers(
         tmp_directory.return_value.__enter__.return_value = '/tmp'
 
         commands = [
+            (rerender, [
```
CJ-Wright: Should we generate these programatically? (every combo of title case, upper case, lower case, comma, colon, semi-colon, etc.)

isuruf: We only need a few tests to see the regex works. I don't see a benefit worth spending time on generating them.

isuruf:
@dougalsutherland, need some help with the issue closed test. I have no experience with using mock. How do I fix it?

conda_forge_webservices/tests/test_commands.py
CJ-Wright: Sorry I didn't realize that was the test, I thought that was the text the bot would accept :sheep:

dougalsutherland:
Your solution there looks good. :)

jakirkham:
LGTM

jakirkham:
If there are no comments, planning on merging this tomorrow after the `conda-smithy` patch release is out.

dougalsutherland:
Noticed a bug (which tests didn't catch because we mock out `rerender`) and a minor phrasing issue. Looks good now.

jakirkham:
Tried to fix some merge conflicts caused by PR ( https://github.com/conda-forge/conda-forge-webservices/pull/186 ). Please feel free to tweak additionally as needed.

jakirkham:
Is this ready to merge, @isuruf?

PR #183, opened 2018-04-19
use newer versions of conda

dougalsutherland:
I don't think this will actually change anything. But might as well update the version of `conda` to 4.5 instead of 4.3, and also verify we have `conda-smithy >=3` rather than just `>=2.4` – just to make sure that no weird constraints end up with us using an old `conda-smithy`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
 
 install:
  - export CONDA_BASE=http://repo.continuum.io/miniconda/Miniconda3
- - wget ${CONDA_BASE}-4.3.30-Linux-x86_64.sh -O miniconda.sh;
+ - wget ${CONDA_BASE}-4.4.10-Linux-x86_64.sh -O miniconda.sh;
  - bash miniconda.sh -b -p $HOME/miniconda
  - export PATH="$HOME/miniconda/bin:$PATH"
  - conda config --set show_channel_urls True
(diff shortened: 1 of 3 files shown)


jakirkham:
So we need a patch release of `conda-smithy` actually as the pinning file is not getting pulled in by default. Am waiting on PR ( https://github.com/conda-forge/conda-smithy/pull/753 ) before doing that.

dougalsutherland:
(Apparently there's not a rerendering test, since these tests passed anyway....)

jakirkham:
Good point. We should start testing that.

dougalsutherland:
Like https://github.com/conda-forge/conda-smithy/pull/754: this should make totally sure that heroku has the latest pinnings.

jakirkham:
New `conda-smithy` is out. So that blocker is resolved.

Not sure if you have seen issue ( https://github.com/conda/conda/issues/6811 ), but would recommend adding a `conda update conda` before doing anything else in these scripts. 4.4.10 suffers from some pretty rough bugs that have since been fixed. So upgrading should get us away from those issues quickly.

Also we probably should be activating `conda`. I know we haven't been, but it's a bad habit that we need to start moving away from `conda` 4.4+ gets weird when one installs things with a non-activated `conda`.

Anything else come to mind @isuruf?

isuruf:
We should have a url that updates conda-smithy and conda-forge-pinning when there's a git webhook event from either of the feedstocks. New conda-smithy's behaviour is to error out if either of them are old. (there's a flag to disable this, but having them updated is better.)

jakirkham:
That's a really great idea. Could we add it to an issue so that we can track it (if there isn't already one)? We might want to couple that with sending out emails when the Heroku deploy succeeds or fails.

dougalsutherland:
Switched to using `conda activate` in both Travis and `run-buildpack`. Think that should work, but someone else should take a look at it. :)

jakirkham:
Thanks @dougalsutherland.

PR #195, opened 2018-05-24
Welcome message for new members

isuruf:


conda_forge_webservices/update_teams.py
```diff
@@ -4,18 +4,78 @@
...
+
+
+def filter_members(members):
+    out = get_filter_out_members()
+    return [m for m in members if m not in out]
+
+
+def get_handles(members):
+    mem = ['@' + m for m in filter_members(members)]
+    return ','.join(mem)
```
jakirkham: Maybe add a space after `,`

jakirkham:
So I've gone back and forth on this. Would be good to hear your thoughts. Namely between commenting on a commit vs. commenting on a PR. It's easier from a programmatic standpoint to do the former, but it is a little less accessible. OTOH it is a little more accessible to comment in PRs, but it is a bit more challenging to approach from a programmatic standpoint. There is the extra caveat that if the change was pushed directly (not in a PR) there is no PR to comment on. This gets a bit more interesting when considering `staged-recipes` additions where the comment may appear either in the feedstock or the staged-recipes PR depending on the approach. Anyways what do you think about all of this?

isuruf:
> It's easier from a programmatic standpoint to do the former, but it is a little less accessible.

Commenting on a commit will appear in the PR, right? I think this is good enough for now and if somebody wants they can change it.

> This gets a bit more interesting when considering staged-recipes additions where the comment may appear either in the feedstock or the staged-recipes PR depending on the approach

Currently, this code is not triggered in the first commit of a feedstock because members are added in the staged-recipes script. I intend to turn that off so that updating team happens only here. (Or we can have staged-recipes script do the commenting on the staged-recipes PR)

isuruf:
I'm going to merge this and play around with it a little bit.

jakirkham:
SGTM. Are you still using that branch or is it ok to remove?

isuruf:
Deleted it

ericdill:
This is awesome! Thanks for the effort here @isuruf and @jakirkham

PR #215, opened 2018-07-26
Update conda_forge_webservices

isuruf:


conda_forge_webservices/update_me.py
```diff
@@ -0,0 +1,40 @@
+import os
+from git import Repo, Actor
+
+from conda_build.conda_interface import (VersionOrder, MatchSpec,
+    get_installed_version, root_dir, get_index, Resolve)
+
+from .utils import tmp_directory
+
+
+def update_me():
```
CJ-Wright: would it be possible to add some docs here? (I guess this should be self evident but...)

dougalsutherland:
Nice. Is the idea that this would just be pinged occasionally, or is there a way to hook it after builds are completed on `master` for conda-forge-pinning / conda-smithy?

isuruf:
> is there a way to hook it after builds are completed on master for conda-forge-pinning / conda-smithy?

Yes, we can look at the circle-ci status.

conda_forge_webservices/update_me.py
isuruf: sure

jakirkham:
Thanks for working on this @isuruf. Was thinking this might be a reasonable way to solve the problem as well.

There is a [Status Event]( https://developer.github.com/v3/activity/events/types/#statusevent ) we could listen for updates too. We may also need to [list the statuses]( https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref ) after and check CircleCI passed. Though that shouldn't be needed if the Status Event has this information already.

PR #224, opened 2018-07-31
Empty commit to rebuild for conda-smithy=3.1.9

CJ-Wright:


CJ-Wright:
This is needed for the new cs release (I think).

CJ-Wright:
@conda-forge/core I'm not certain why the tests are failing.

mwcraig:
I think it is because the token (stored as an encrypted variable in `.travis.yml`) is not available for PRs, so anything that needs the token `$GH_TOKEN` is failing.

CJ-Wright:
Ah ok. In that case could I get merge/review?

dougalsutherland:
i don't know that we really need review for a non-change. :)

PR #235, opened 2018-10-13
Remove checking for conda-build updates for now

isuruf:
It looks like conda doesn't like updating to the latest and this results in dozens of empty commits for no reason.

### DIFF
--- a/conda_forge_webservices/update_me.py
+++ b/conda_forge_webservices/update_me.py
@@ -11,7 +11,7 @@ def update_me():
     """
     Update the webservice on Heroku by pushing a commit to this repo.
     """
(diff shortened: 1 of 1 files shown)


jakirkham:
Looking into it a bit, I think what is happening is we are always updating any time there is a push event (though maybe nothing was deployed) and a status event (though it may not have passed or may not be the one we care about). Have added PR ( https://github.com/conda-forge/conda-forge-webservices/pull/236 ), which checks the combined status to see if all checks have passed before deploying. There are some tradeoffs, but it should cutdown on the extra updates. Would be good to hear your thoughts on it.

PR #236, opened 2018-10-13
Check the combined status before updating

jakirkham:
Alternative to PR ( https://github.com/conda-forge/conda-forge-webservices/pull/235 ).

Only updates the webservice if the combined status is a pass. This should avoid multiple updates to the webservice when the package is not there. Though it may mean the specific architecture and Python version that we need has already been deployed for a little bit.

cc @isuruf

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -25,6 +25,17 @@ import conda_forge_webservices.commands as commands
 import conda_forge_webservices.update_me as update_me
 
 
+def get_combined_status(token, repo_name, sha):
+    # Get the combined status for the repo and sha given.
+
+    gh = github.Github(token)
+    repo = gh.get_repo(repo_slug)
+    commit = repo.get_commit(sha)
+    status = commit.get_combined_status()
+
+    return status.state
+
+
 def print_rate_limiting_info_for_token(token, user):
     # Compute some info about our GitHub API Rate Limit.
     # Note that it doesn't count against our limit to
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/webapp.py
```diff
@@ -268,10 +279,25 @@ def post(self):
...
+                token = os.environ.get('GH_TOKEN')
+                state = get_combined_status(token, repo_name, sha)
+
+            # Update if the combined status is a success
+            if state == 'success':
+                update_me.update_me()
+
             print_rate_limiting_info()
-        else:
+        elif event != 'push':
```
jakirkham: Maybe this should just be disabled on the webhooks. Thoughts?

isuruf:
The way I've written the webhook is that it checks the versions of the conda packages in its environment against the latest versions available in anaconda.org. So, it doesn't matter whether we check the combined status or just any status.

jakirkham:
Not sure I follow. If one looks at the commits on `master` recently, one can see there are a bunch of commits to update `conda-build`. It would be better to have just one. That's what this PR is solving.

conda_forge_webservices/webapp.py
scopatz: Seems fine.

jakirkham:
Thoughts on this @conda-forge/core?

Recently (past 30 mins) we had 15 commits pushed to webservices to update the `conda-build` version. This should reduce it to 1.

isuruf:
See my comment.

> So, it doesn't matter whether we check the combined status or just any status.

Having the 15 commits some other issue.

PR #248, opened 2018-12-30
Unpin conda-build 3.16

jakirkham:
This was pinned due to a change in conda-build 3.17 that broke conda-smithy. Since then conda-smithy has been fixed to work with both 3.16 and 3.17. Hence we can relax this pin.

xref: https://github.com/conda-forge/conda-smithy/pull/962


jakirkham:
cc @conda-forge/core

PR #253, opened 2019-02-13
trying to make the rerender command not fail silently

richardjgowers:
I ran into a weird state where 'conda forge please rerender' would say that nothing needed doing, when in fact it had actually just failed (and therefore didn't see a change in the repo).

I think by checking the return code of the subprocess call you can at least see when this happens and advise.

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -118,7 +118,7 @@ def pr_detailed_comment(org_name, repo_name, pr_owner, pr_repo, pr_branch, pr_nu
                     message = textwrap.dedent("""
                         Hi! This is the friendly automated conda-forge-webservice.
 
-                        I tried to {} for you but ran into some issues, try doing these actions locally instead.
+                        I tried to {} for you but ran into some issues, please ping conda-forge/core for further assistance.
                         """)
                 else:
                     message = textwrap.dedent("""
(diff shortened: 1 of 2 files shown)


conda_forge_webservices/commands.py
```diff
@@ -110,11 +114,18 @@ def pr_detailed_comment(org_name, repo_name, pr_owner, pr_repo, pr_branch, pr_nu
...
             else:
-                message = textwrap.dedent("""
-                    Hi! This is the friendly automated conda-forge-webservice.
+                if rerender_error:
+                    message = textwrap.dedent("""
+                        Hi! This is the friendly automated conda-forge-webservice.
 
-                    I tried to {} for you, but it looks like there was nothing to do.
-                    """).format(changes_str)
+                        I tried to {} for you but ran into some issues, try doing these actions locally instead.
```
jakirkham: Would actually change this to suggest pinging core instead.

jakirkham:
Thanks for working on this @richardjgowers! Added a small suggestion above.

conda_forge_webservices/commands.py
richardjgowers: As in the bot auto pinging core, or suggesting user to ping core?

richardjgowers: ok done!

scopatz:
I think the CI failures are normal here, right?

mariusvniekerk:
yeah don't think the forks have access to the token

jakirkham:
It’s up to us to push the change onto a non-`master` branch to get the tests to run.

isuruf:
Yes, this PR is failing the build on master.

PR #300, opened 2020-02-01
Move to org wide hooks

isuruf:
Checklist
* [x] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -130,7 +130,7 @@ class LintingHookHandler(tornado.web.RequestHandler):
                     label['name'] == 'stale'
                     for label in body['pull_request']['labels']
                 )
-            elif not repo.endswith("-feedstock"):
+            elif not repo_name.endswith("-feedstock"):
                 self.set_status(404)
                 self.write_error(404)
                 return
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -211,7 +211,7 @@ class UpdateTeamHookHandler(tornado.web.RequestHandler):
             if 'head_commit' in body:
                 commit = body['head_commit']['id']
             # Only do anything if we are working with conda-forge, and a push to master.
-            if owner == 'conda-forge' and repo.endswith("-feedstock") and ref == "refs/heads/master":
+            if owner == 'conda-forge' and repo_name.endswith("-feedstock") and ref == "refs/heads/master":
                 update_teams.update_team(owner, repo_name, commit)
             print_rate_limiting_info()
         else:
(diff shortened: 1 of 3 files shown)


conda_forge_webservices/webapp.py
```diff
@@ -207,7 +211,7 @@ def post(self):
             if 'head_commit' in body:
                 commit = body['head_commit']['id']
             # Only do anything if we are working with conda-forge, and a push to master.
-            if owner == 'conda-forge' and ref == "refs/heads/master":
+            if owner == 'conda-forge' and repo.endswith("-feedstock") and ref == "refs/heads/master":
```
jakirkham: Why limit it to `master`?

isuruf: This was there before. (This hook updates the team if master changes and we only look at master for updating the team)

isuruf:
Well, I was careful. A thorough review is appreciated.

jakirkham:
So I'm probably missing context here, but what will get by using the org webhook? Some events we don't get now? Easier to work with requests? Something else?

beckermr:
Fewer api requests when making feedstocks.

jakirkham:
Do we have a sense of how many requests we would save?

conda_forge_webservices/webapp.py
beckermr: This should be `repo_name` as opposed to `repo`, right?

conda_forge_webservices/webapp.py
```diff
@@ -230,7 +234,7 @@ def post(self):
             repo_name = body['repository']['name']
             owner = body['repository']['owner']['login']
             # Only do anything if we are working with conda-forge
-            if owner != 'conda-forge':
+            if owner != 'conda-forge' or not (repo == "staged-recipes" or repo.endswith("-feedstock")):
```
beckermr: ditto here I think

conda_forge_webservices/webapp.py
```diff
@@ -257,7 +261,7 @@ def post(self):
             issue_num = body['issue']['number']
 
             # Only do anything if we are working with conda-forge
-            if owner != 'conda-forge':
+            if owner != 'conda-forge' or not (repo == "staged-recipes" or repo.endswith("-feedstock")):
```
beckermr: and here

isuruf:
Added more tests so that bugs like before will be tested.

beckermr:
Also it may be good to run a branch in staged recipes to slowly remove the old hooks.

isuruf:
> We’ll need to add the new webhook a before merging

I already did

isuruf:
Thanks @beckermr

isuruf:
@jakirkham, forgot to answer your comment. staged-recipes is now failing at the stage where webhooks are added. For eg: https://travis-ci.com/conda-forge/staged-recipes/builds/147001015
This is one step towards fixing that.

PR #303, opened 2020-02-02
ENH add cron updates

### LINKED ISSUE #256: update me script is not working anymore.
Probably because of https://github.com/conda-forge/conda-forge-webservices/blob/master/conda_forge_webservices/webapp.py#L32

### LINKED ISSUE #274: NameError: name 'repo_slug' is not defined
Caught the following error from the Heroku logs. Posting it here.

```python
2019-10-11T16:39:33.587511+00:00 app[web.1]: ERROR:tornado.application:Uncaught exception POST /conda-webservice-update/hook (140.82.115.250)
2019-10-11T16:39:33.587515+00:00 app[web.1]: HTTPServerRequest(protocol='http', host='conda-forge.herokuapp.com', method='POST', uri='/conda-webservice-update/hook', version='HTTP/1.1', remote_ip='140.82.115.250')
2019-10-11T16:39:33.587518+00:00 app[web.1]: Traceback (most recent call last):
2019-10-11T16:39:33.587520+00:00 app[web.1]:   File "/app/.conda/lib/python3.7/site-packages/tornado/web.py", line 1697, in _execute
2019-10-11T16:39:33.587523+00:00 app[web.1]:     result = method(*self.path_args, **self.path_kwargs)
2019-10-11T16:39:33.587525+00:00 app[web.1]:   File "/app/conda_forge_webservices/webapp.py", line 294, in post
2019-10-11T16:39:33.587528+00:00 app[web.1]:     state = get_combined_status(token, repo_name, sha)
2019-10-11T16:39:33.587530+00:00 app[web.1]:   File "/app/conda_forge_webservices/webapp.py", line 32, in get_combined_status
2019-10-11T16:39:33.587532+00:00 app[web.1]:     repo = gh.get_repo(repo_slug)
2019-10-11T16:39:33.587535+00:00 app[web.1]: NameError: name 'repo_slug' is not defined
2019-10-11T16:39:33.590722+00:00 app[web.1]: ERROR:tornado.access:500 POST /conda-webservice-update/hook (140.82.115.250) 143.24ms
2019-10-11T16:39:33.580145+00:00 heroku[router]: at=info method=POST path="/conda-forge-command/hook" host=conda-forge.herokuapp.com request_id=0d16f224-6281-4b5c-a97d-0f1f8ccdbf3e fwd="140.82.115.251" dyno=web.1 connect=1ms service=1196ms status=200 bytes=163 protocol=http
```

### LINKED ISSUE #299: Add an hourly cron job to ping the `update_me` URL
This would make the rerendering failures due to new conda-smithy or conda-forge-pinning go away inside an hour.

beckermr:
This PR adds cron updates to make sure the app is updated hourly. 

Closes #299 
Closes #256 
Closes #274 

Checklist
* [x] CI passed on the branch


beckermr:
@isuruf I could not think of great tests for this PR. If you have ideas, i'm happy to implement them!

conda_forge_webservices/webapp.py
```diff
@@ -24,16 +24,8 @@
...
-    # Get the combined status for the repo and sha given.
-
-    gh = github.Github(token)
-    repo = gh.get_repo(repo_slug)
-    commit = repo.get_commit(sha)
-    status = commit.get_combined_status()
-
-    return status.state
+UPDATED_ME_DELAY = 55 * 60  # 55 mins
+UPDATED_ME_LAST = datetime.now().astimezone(pytz.UTC)
```
isuruf: These won't work if there are threads involved right?

beckermr: Each thread in the tornado server is actually a separate python process. So when each process starts, it will set the time it started. Then when a request to this endpoint comes in, it will get routed to only one of the processes. That process will compare the time, and if it has been long enough, it will push a commit which will end up rebooting the entire server. Then it all starts over. 

So I think this works fine.

isuruf: If they are different python processes these variables are not global right? They are thread local. Therefore I don't see any advantage to having this code. Maybe we could check the last git commit's time?

beckermr: Yeah so I’m not claiming this way has some advantage over looking at the commit time or anything. Both the time of the last commit and this way are subject to race conditions because we are not obtaining a real lock over the processes. 

I’m happy to switch to the time of the last commit but this way was the first thing that came to mind.

beckermr: OTOH the check on the versions of conda, conda build, and conda smithy will very likely prevent the abuse mechanism I am thinking of.

beckermr: Actually no, I like the code as-is. This way of checking the time doesn't involve pulling down the repo or doing any solving which will keep the server less busy. Both ways have race conditions, so this seems better.

beckermr: Ohhhhh! I had one more idea! We could push a random hash to a file and then if git rejects us, some other process in the server did it first. This would be more expensive but would probably be the safest thing. Thoughts?

beckermr: We could also use a random hash in the header of the request, with the hash value living on circleci and heroku in their env. This is probably the best way.

beckermr:
OK @isuruf. I switched to a key-based system. We'll need to add the key to both heroku and circle.

isuruf:
Thanks @beckermr. I now understand what you were trying to do. You want to avoid somebody abusing this. (I was thinking about a case where we had multiple commits because conda wasn't installing the latest conda-build, but that's a different thing we can handle later)

beckermr:
Alright! LGTM for the token based system?

isuruf:
+1

PR #338, opened 2020-03-05
PERF background slow tasks

### LINKED ISSUE #25: Background the linting process
As per the suggestion at https://devcenter.heroku.com/articles/background-jobs-queueing we probably shouldn't be blocking the webservice for the few seconds that it takes to run the linting.

beckermr:
Closes #25 

Checklist
* [ ] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -163,6 +163,7 @@ class LintingHookHandler(tornado.web.RequestHandler):
                 print("===================================================")
 
                 lint_info = await tornado.ioloop.IOLoop.current().run_in_executor(
+                    None,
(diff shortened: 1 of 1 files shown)


jakirkham:
Do we need to await these somewhere or is that already done for us?

beckermr:
They are awaited right in the function with the await keywords

jakirkham:
SGTM. Merge when ready 🙂

beckermr:
See this one: http://www.tornadoweb.org/en/stable/faq.html#why-isn-t-this-example-with-time-sleep-running-in-parallel

beckermr:
See item 3

PR #339, opened 2020-03-06
ENH try a thread pool

beckermr:
Checklist
* [x] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -6,8 +6,9 @@ import tornado.web
 import hmac
 import hashlib
 import json
+from concurrent.futures import ThreadPoolExecutor
+import atexit
 
-import requests
 import github
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/webapp.py
```diff
@@ -18,6 +19,25 @@
 from conda_forge_webservices.update_me import get_current_versions
 
 
+POOL = None
+
+
+def _thread_pool():
+    global POOL
+    if POOL is None:
+        POOL = ThreadPoolExecutor(max_workers=8)
```
jakirkham: How did we determine `8`? Also is there a way to query this from the runtime environment?

beckermr: So Tornado launches one process per WEB_CONCURRENCY which is set by heroku for the env to correspond to the number of CPUs roughly. This code thus launches 8 threads per CPU core. I am guessing at this number. Since most of the operations in the code are I/O bound, we should be able to handle more than one thread per core just fine and in my experience with other I/O bound codes, 8 was OK. We'll see if it is too much.

beckermr: The only reason I am trying this is that the current code is growing memory usage at a steady clip so I'd like to test if this is better and doesn't do that.

jakirkham: Sure. Though wouldn't each thread add to the memory usage?

beckermr: depends

beckermr: this appears to have solved the memory issues on the heroku instance

cc @isuruf

PR #352, opened 2020-03-21
REF don't make as many statuses

beckermr:
Right now we make what appears to be more than one status per commit. This PR puts in a check to make sure only make a new status if it is different or we have not done it before.

Checklist
* [x] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -225,6 +225,7 @@ def set_pr_status(owner, repo_name, lint_info, target_url=None):
         commit = repo.get_commit(lint_info['sha'])
 
         # get the last github status by the linter, if any
+        # API emmits these in reverse time order so first is latest
         statuses = commit.get_statuses()
         last_state = None
         for status in statuses:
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/linting.py
```diff
@@ -223,16 +223,34 @@ def set_pr_status(owner, repo_name, lint_info, target_url=None):
...
+            if status.context == "conda-forge-linter":
+                last_state = status.state
+
+        # convert the linter status to a state
+        lint_status_to_state = {"good": "success", "mixed": "success"}
+        lint_last_state = lint_status_to_state.get(lint_info['status'], "failure")
+
+        # make a status only if it is different or we have not ever done it
+        # for this commit
+        if last_state is None or last_state != lint_last_state:
```
isuruf: This should check that the target URLs are different as well.

beckermr: Why?

isuruf: For eg: if the commit had two bad statuses.
1. lint failure due to smithy failure - comment 1
2. actual lint - comment 2

Both comments have the same state, but the target_url is different.

beckermr: Ohhhh. A bad status should point back to the most recent comment with the correct lints. Will fix.

jakirkham:
Thanks Matt! 😀

PR #356, opened 2020-04-10
ENH endpoints for cfep-13

beckermr:
Checklist
* [x] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/feedstock_outputs.py
+++ b/conda_forge_webservices/feedstock_outputs.py
@@ -33,29 +33,45 @@ def register_feedstock_token_handler(feedstock):
 
     with tempfile.TemporaryDirectory() as tmpdir:
         with pushd(tmpdir):
-            subprocess.run(
-                ["git", "clone", "--depth=1", feedstock_url],
-                check=True,
-            )
+            try:
+                subprocess.run(
+                    ["git", "clone", "--depth=1", feedstock_url],
+                    check=True,
+                )
+            except subprocess.CalledProcessError:
+                print("    could not clone the feedstock")
+                error = True
 
-            with pushd(feedstock_name):
-                try:
-                    subprocess.run(
-                        ["conda-smithy", "generate-feedstock-token"],
-                        check=True,
-                    )
-                    subprocess.run(
-                        ["conda-smithy", "register-feedstock-token"],
-                        check=True,
-                    )
-                finally:
+            if not error:
+                with pushd(feedstock_name + "-feedstock"):
                     try:
-                        os.remove(os.path.expanduser(
-                            "~/.conda_smithy/conda-forge_"
-                            "%s_feedstock.token" % feedstock_name
-                        ))
-                    except Exception:
-                        error = True
+                        try:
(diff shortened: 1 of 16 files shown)


beckermr:
@isuruf @conda-forge/core I'd love some initial code review here if you have the time. 

I have added three endpoints.

1. feedstock-tokens/registration: The intent is that the registration endpoint can be used by both staged recipes and the admin bot to register and provision feedstock tokens. If a token already exists, it will fail. It requires the feedstock token for staged recipes.

2. feedstock-outputs/validate: On PRs in feedstocks, we will call the validate endpoint to make sure the outputs are valid. It does not require a token to use (so it can be called from forks and staged recipes PRs) and accepts a simple list of the outputs in the request body. It returns a dictionary mapping those outputs to true or false if they are allowed for a feedstock. If this fails, the PR will fail, indicating an overlapping output.

3. feedstock-outputs/copy: Once a PR is merged to master, the CI scripts will call the copy endpoint. This one requires a valid feedstock token. It returns a dictionary of any outputs that were valid mapped to whether or not they were copied. We can use this to mark PRs on master as failed if the outputs never make it to the main channel post staging.

beckermr:
@conda-forge/core this one is ready for review

beckermr:
Just an FYI, I put some tests that run against a local instance of the webserver in the scripts dir. These could be run in the CI service, but they involve live calls to the various anaconda APIs, running conda-build etc, so I think for now I will leave them out.

beckermr:
bump here @conda-forge/core

conda_forge_webservices/feedstock_outputs.py
```diff
@@ -6,67 +6,184 @@
...
+        feedstock_name = feedstock
+
+    feedstock_url = "https://github.com/conda-forge/%s-feedstock.git" % feedstock_name
+    token_path = os.path.expanduser(
+        "~/.conda-smithy/conda-forge_%s_feedstock.token" % feedstock_name
+    )
+
+    try:
+        with tempfile.TemporaryDirectory() as tmpdir:
+            with pushd(tmpdir):
```
scopatz, line 58: These can be combined to the same with statement, and save an indentation level

conda_forge_webservices/feedstock_outputs.py
```diff
@@ -6,67 +6,184 @@
...
+                except subprocess.CalledProcessError:
+                    print("    could not clone the feedstock")
+                    return True
+
+                with pushd(feedstock_name + "-feedstock"):
+                    try:
+                        subprocess.run(
+                            ["conda-smithy", "generate-feedstock-token"],
+                            check=True,
+                        )
```
scopatz, line 70: It would probably be good to make a `safe_run()` function, so that there aren't all these bits of repeated code

conda_forge_webservices/feedstock_outputs.py
beckermr: 🤦 I keep forgetting this!

scopatz:
Thanks @beckermr!

PR #373, opened 2020-04-28
ENH add appveyor is ok list handling

beckermr:
This PR implements the appveyor is ok list.

We do the following

1. For any feedstock in the whitelist, they can make copy requests without a valid feedstock token.
2. These requests cannot create new outputs that have not already been registered.
3. These requests only work on win-64.

Checklist
* [x] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/feedstock_outputs.py
+++ b/conda_forge_webservices/feedstock_outputs.py
@@ -256,7 +256,9 @@ def _is_valid_output_hash(outputs):
     return valid
 
 
-def is_valid_feedstock_output(project, outputs, register=True):
+def is_valid_feedstock_output(
+    project, outputs, register=True, must_explicitly_exist=False
+):
     """Test if feedstock outputs are valid (i.e., the outputs are allowed for that
     feedstock). Optionally register them if they do not exist.
 
@@ -268,9 +270,13 @@ def is_valid_feedstock_output(project, outputs, register=True):
         A list of ouputs top validate. The list entries should be the
         full names with the platform directory, version/build info, and file extension
         (e.g., `noarch/blah-fa31b0-2020.04.13.15.54.07-py_0.tar.bz2`).
-    register : bool
+    register : bool, optional
         If True, attempt to register any outputs that do not exist by pushing
         the proper json blob to `output_repo`. Default is True.
+    must_explicitly_exist : bool, optional
+        If True, the output must be already registered and exist for it to be
+        valid. This option is used for appveyor-only uploads where we cannot
+        verify the request.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


beckermr:
OK @conda-forge/core @jjhelmus @isuruf I'd appreciate a close look at the logic in this PR. See the details above. Right now python is the only feedstock on this list.

conda_forge_webservices/feedstock_outputs.py
```diff
@@ -269,17 +270,21 @@ def is_valid_feedstock_output(project, outputs, register=True):
...
+        verify the request.
 
     Returns
     -------
     valid : dict
         A dict keyed on output name with True if it is valid and False
         otherwise.
     """
-    feedstock = project.replace("-feedstock", "")
+    feedstock = project[:-len("-feedstock")]
```
scopatz, line 287: What if `project` doesn't end in "-feedstock"?

conda_forge_webservices/tests/test_feedstock_outputs.py
```diff
@@ -216,15 +241,17 @@ def test_is_valid_output_hash():
     }
 
 
+@pytest.mark.parametrize("must_explicitly_exist", [True, False])
 @pytest.mark.parametrize("register", [True, False])
 @pytest.mark.parametrize(
-    "project", ["foo", "foo-feedstock", "blah", "blarg", "boo"]
+    "project", ["foo-feedstock", "blah-feedstock", "blarg-feedstock", "boo-feedstock"]
```
scopatz, line 247: I think we should keep the previous capability

beckermr: It was removed from smithy. We now require the full repo name.

conda_forge_webservices/feedstock_outputs.py
beckermr: This code assumes that it does. I will put an assert above to make sure it fails.

conda_forge_webservices/tests/test_feedstock_outputs.py
beckermr: wait no sorry. I was thinking of tokens.

beckermr: I removed this because it got confusing when I was or was not using the repo name versus the one that was munged. Plus doing things like `.replace("-feedstock", "")` is dangerous since a feedstock could be called `my-feedstock-wrangler-feedtsock` which might generate the wrong outputs and misregister things.

Thus I changed the code to alwasy use the full repo name on input.

conda_forge_webservices/feedstock_outputs.py
beckermr, line 287:
```suggestion
    assert feedstock.endswith("-feedstock")
    feedstock = project[:-len("-feedstock")]
```

conda_forge_webservices/tests/test_feedstock_outputs.py
beckermr: For tokens, I made the same change FWIW.

conda_forge_webservices/webapp.py
```diff
@@ -534,19 +547,33 @@ async def post(self):
             or feedstock is None
             or outputs is None
             or channel is None
```
beckermr, line 549:
```suggestion
            or channel is None
            or not _repo_exists(feedstock)
```

beckermr:
OK @scopatz I put back the code you wanted and fixed up the logic. Ready for another round!

beckermr:
We good to merge here @scopatz ?

PR #409, opened 2020-08-27
Remove timestamp

isuruf:
Checklist
* [ ] CI passed on the branch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,9 @@ ARG CONDA_DIR="/opt/conda"
 ENV PATH="$CONDA_DIR/bin:$PATH"
 ENV PYTHONDONTWRITEBYTECODE=1
 
+# bust the docker cache so that we always rerun the installs below
+ADD http://www.randomtext.me/api/gibberish /opt/docker/etc/gibberish
+
 # Install conda
 COPY conda-requirements.txt /
 RUN echo "**** install dev packages ****" && \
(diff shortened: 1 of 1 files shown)


isuruf:
Are you sure? If so, can you propose a new website? The URL doesn't work

beckermr:
I am pretty sure. Let me google around.

beckermr:
see this from the docker docs

> Aside from the ADD and COPY commands, cache checking does not look at the files in the container to determine a cache match. For example, when processing a RUN apt-get -y update command the files updated in the container are not examined to determine if a cache hit exists. In that case just the command string itself is used to find a match.

Dockerfile
```diff
@@ -38,9 +38,6 @@ ARG CONDA_DIR="/opt/conda"
 ENV PATH="$CONDA_DIR/bin:$PATH"
 ENV PYTHONDONTWRITEBYTECODE=1
 
-# make sure the install below is not cached by docker
-ADD http://worldclockapi.com/api/json/utc/now /opt/docker/etc/timestamp
-
 # Install conda
```
beckermr, line 41:
```suggestion
# bust the docker cache so that we always rerun the installs below
ADD http://www.randomtext.me/api/gibberish /opt/docker/etc/gibberish

# Install conda
```

beckermr: @isuruf - try this

PR #413, opened 2020-11-26
ENH try to make the package before the copy to get attrs

beckermr:
This PR tries to make the package on the production server before we copy it. This should allow us to properly set the package attributes.

Checklist
* [x] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/feedstock_outputs.py
+++ b/conda_forge_webservices/feedstock_outputs.py
@@ -99,7 +99,7 @@ def _make_package(name, ac_prod):
         "description": pkg_info["description"],
         "license": pkg_info["license"],
         "license_url": pkg_info["license_url"],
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks for working on this Matt! 😄

Do you recall what is still incomplete here (thinking about the draft status)? 🙂

beckermr:
This won't fix the issue. The copy removes the info on anaconda.org anyways. :/

jakirkham:
Oh interesting. What if we add the info after the copy? Or are there issues with that approach as well?

beckermr:
We cannot adjust the info on anaconda.org after we create the package. We will have to be able to do this to fix the packages we already have.

beckermr:
Closing as this has been fixed upstream.

PR #546, opened 2023-03-07
Add `shellcheck` to requirements

jakirkham:
To provide for `shellcheck`-based linting, install it in webservices.

<hr>

Checklist
* [ ] CI passed on the branch

### DIFF
--- a/conda-requirements.txt
+++ b/conda-requirements.txt
@@ -23,6 +23,7 @@ pyyaml
 requests
 ruamel.yaml>=0.16
 ruamel.yaml.jinja2
+shellcheck
 tini
 tornado
 tqdm


beckermr:
I think this will force shellcheck linking as required in all feedstocks. Is that the policy we want?

jakirkham:
No [the default]( https://github.com/conda-forge/conda-smithy/blob/4837b90f8194559f1df3a96de1113720f514858a/conda_smithy/lint_recipe.py#L682 ) is [`shellcheck` is disabled]( https://github.com/conda-forge/conda-smithy/blob/4837b90f8194559f1df3a96de1113720f514858a/conda_smithy/lint_recipe.py#L702-L704 ). Users need to [opt-in]( https://github.com/conda-forge/conda-smithy/blob/4837b90f8194559f1df3a96de1113720f514858a/tests/conda-forge.yml#L1-L2 ).

Though atm they can't do that since we don't have `shellcheck` installed in the webservice. So was trying to fix that.

The next step is documenting it, which PR ( https://github.com/conda-forge/conda-forge.github.io/pull/1910 ) would address. Though would want to make sure it works before suggesting it as an option.

jakirkham:
Thanks Matt! 🙏

PR #579, opened 2024-03-21
ENH update webservices dispatch action docker image for new releases

### LINKED ISSUE #578: rebuild docker images when versions change
Right now we redeploy the webservices when smithy gets an update so that we get the latest linting. We should use a repo dispatch event to rebuild our docker images for automerge and rerendering using the same trigger. We'll need to add the pinnings repo, conda-build, boa, conda, and mamba to the list of packages used to trigger an update.

beckermr:
Checklist
* [x] CI passed on the branch

closes #578

### DIFF
--- a/conda_forge_webservices/update_me.py
+++ b/conda_forge_webservices/update_me.py
@@ -4,6 +4,7 @@ import json
 import tempfile
 import shutil
 import logging
+import github
 
 from git import Repo
 import requests
@@ -97,3 +98,11 @@ def main():
         finally:
             if tmpdir is not None:
                 shutil.rmtree(tmpdir)
+
+        try:
+            gh = github.Github(get_app_token_for_webservices_only())
+            repo = gh.get_repo("conda-forge/webservices-dispatch-action")
+            workflow = repo.get_workflow("tests.yml")
+            workflow.create_dispatch("main")
+        except Exception as e:
+            print(f"workflow_dispatch for webservices-dispatch-action failed: {e}")
--- a/conda_forge_webservices/update_me.py
+++ b/conda_forge_webservices/update_me.py
@@ -15,7 +15,7 @@ from conda_forge_webservices.tokens import get_app_token_for_webservices_only
 
 LOGGER = logging.getLogger("conda_forge_webservices.update_me")
 
-PKGS = ["conda-smithy", "conda", "conda-build"]
+PKGS = ["conda-smithy", "conda", "conda-build", "conda-libmamba-solver", "mamba"]
 
 
 def _run_git_command(args):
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/update_me.py
```diff
@@ -14,7 +15,7 @@
 
 LOGGER = logging.getLogger("conda_forge_webservices.update_me")
 
-PKGS = ["conda-smithy"]
+PKGS = ["conda-smithy", "conda", "conda-build"]
```
jakirkham, line 18: Would it make sense to add

```suggestion
PKGS = ["conda-smithy", "conda", "conda-build", "conda-libmamba-solver", "mamba"]
```

jakirkham (commented):
Thanks Matt! 🙏

Had a question below

conda_forge_webservices/update_me.py
beckermr: yes and thank you!

jakirkham (approved):

mbargull (approved):

jakirkham:
Rerunning CI as there was an error

jakirkham:
Thanks all! 🙏

jakirkham:
Guessing this looks as expected: https://github.com/conda-forge/conda-forge-webservices/commit/8ccbbe9d2beaa6fabeda23a4b2a5b6f72f9e34d9

isuruf:
This updates the heroku webservice too. There should be two package lists. We had a discussion somewhere (can't remember where) about only doing conda-smithy to avoid frequent restarts of the webserver resulting in requests getting dropped.

beckermr:
The rate of drops from this is small and historically we had updated the webservices for other things IIRC.

isuruf:
But why merge these two completely different things into one thing? I understand why we need to update the docker image. but I don't see why we need to restart the heroku server for these updates. Is it because it was easier to write the code?

beckermr:
The docker image is tied to the webservices code and is an extension of it. We update the two in tandem because to ensure consistency.

isuruf:
> The docker image is tied to the webservices code and is an extension of it. 

How? The docker image is for the automerge/rerender action and the webservices code is for the service that runs on heroku right?

beckermr:
The docker image is used for rerendering and version updates. 

Automerge is done via a different docker image. 

The webservices passes specific json blobs to the docker image via the repo dispatch gh events. The docker image relies on the formats of these blobs.

isuruf:
From the title of this PR and the linked issue, the issue was that updates in each rerender call was costly and keeping the docker image up-to-date was what this PR intended to do right? I don't understand the additional reasons you are saying about consistency. We never had these mysterious issues. Can you create an issue about these mysterious issues?

beckermr:
I'm not saying anything mysterious I don't think. You are correct that we are using this cron job to update two things that happen to need updates for the same reasons. 

The webservices GitHub action doesn't operate without the heroku service bumping it via the GitHub api. It really is an extension of the webservices code base. In fact we used to rerender right in the heroku server. 

For this reason, I don't think it is so bad to update them both at once triggered on the same event. 

In any case, I don't actually care so do feel free to change things to how you want them. We'll need to bump the image no matter how it gets done.

PR #580, opened 2024-03-21
Update to latest Miniforge3

jakirkham:
* Update to latest installer
* Switch to Miniforge3 (as Mambaforge is deprecated)
* Use `sha256` checksums
* Template installer version (makes it easier to update version and checksum in the future)
* Use `conda install` again (now using `libmamba` solver by default)
* Update messaging and naming to refer to Miniforge3 consistently

<hr>

Checklist
* [x] Pushed the branch to main repo
* [x] CI passed on the branch

PR #585, opened 2024-03-22
Use Mamba install

jakirkham:
Heroku is running out of memory during the build. Switch back to Mamba for the install to see if that helps.

Checklist
* [ ] CI passed on the branch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,7 @@ RUN echo "**** install dev packages ****" && \
     conda config --show-sources  && \
     conda config --set always_yes yes && \
     conda config --set solver libmamba && \
-    conda install --quiet --file conda-requirements.txt && \
+    mamba install --quiet --file conda-requirements.txt && \
     conda clean --all --force-pkgs-dirs --yes && \
     find "$CONDA_DIR" -follow -type f \( -iname '*.a' -o -iname '*.pyc' -o -iname '*.js.map' \) -delete && \
     \


jakirkham:
Here is [the Heroku log]( https://dashboard.heroku.com/apps/conda-forge/activity/builds/4c100ee3-7aaf-4ca4-aa4a-5fcb8b915623 )

jakirkham:
@beckermr success! 🎉

jakirkham:
Testing a re-render to make sure things are working: https://github.com/conda-forge/ucx-split-feedstock/issues/162

jakirkham:
Seems to be working

mbargull:
FYI, you most likely ran into https://github.com/conda/conda/issues/13541#issue-2108238514 which caused a sharp memory usage spike with the latter being avoided in `conda>=24.3` via https://github.com/conda/conda/pull/13628 .
But even with `conda=24.1` you'd probably not run into that issue -- for `conda-forge` at least -- since the code for `repodata.json.zst` support does not take that problematic code path.

IOW, the issue was most likely due to having `conda<=23.11` in our installers and once we update to `conda>=24.1` (or better, `conda>=24.3`), the peak memory usage should be about the same whether `conda` or `mamba` is used.

mbargull:
Yes, but no particular rush, IMO. The bug mentioned above has been present in `conda>=4.12,<24.3`, so not a particular new thing ;).

jakirkham:
Explains why we didn't see it before. This was using 4.11.0 previously (with Mamba)

We just updated ( https://github.com/conda-forge/conda-forge-webservices/pull/580 ) 😬

A new release would be welcome, but agree not a rush. We have cobbled together a working solution to tide us over until then

jakirkham:
Started updating Miniforge in a follow up PR ( https://github.com/conda-forge/conda-forge-webservices/pull/598 ). Though maybe we can go to 24.3 if that is ready ( https://github.com/conda-forge/miniforge/pull/574 )

jakirkham:
We have now upgraded to Miniforge 24.3.0 and are using Conda again for installs. Thanks all! 🙏

PR #595, opened 2024-04-02
Simply use ruamel.yaml for YAML parsing

jakirkham:
Rewrite logic that was using PyYAML to use ruamel.yaml as is used elsewhere in the code. This eliminates the need for PyYAML as a dependency.

<hr>

Checklist
* [x] CI passed on the branch

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -1,5 +1,6 @@
 from git import GitCommandError, Repo, Actor
 import github
+import io
 import os
 import re
 import time
@@ -9,7 +10,6 @@ from ruamel.yaml import YAML
 import requests
 from requests.exceptions import RequestException
 import logging
-import yaml
 
 # from .utils import tmp_directory
 from .linting import compute_lint_message, comment_on_pr, set_pr_status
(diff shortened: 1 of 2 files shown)


jakirkham:
@beckermr could you also please review this PR?

conda_forge_webservices/commands.py
```diff
@@ -643,19 +643,23 @@ def add_user(repo, user):
 
     recipe_path = os.path.join(repo.working_dir, "recipe", "meta.yaml")
     co_path = os.path.join(repo.working_dir, ".github", "CODEOWNERS")
+    yaml = YAML()
```
beckermr, line 646:
```suggestion
    yaml = YAML(typ="safe")
```

We should be careful since the recipe is from users.

beckermr (changes requested):

conda_forge_webservices/commands.py
jakirkham: Sounds reasonable

We do have a bunch of places where this is not set now. Maybe it's worth filing an issue to fix those up too (not sure if those will be as simple without looking more deeply)?

beckermr: https://github.com/conda-forge/conda-forge-webservices/issues/600

PR #609, opened 2024-04-03
Add GHA run link to commit messages

jaimergp:
Checklist
* [X] Pushed the branch to main repo
* [ ] CI passed on the branch

Adds the GHA run to all commit messages produced during tests (or any other context where `$ACTION_URL` has been defined).


beckermr (changes requested):
Awesome!

You missed a few in these files

 - https://github.com/conda-forge/conda-forge-webservices/blob/link-to-gha-run/scripts/test_cfep13_copy.py
 - https://github.com/conda-forge/conda-forge-webservices/blob/link-to-gha-run/scripts/test_cfep13_endpoints.py

beckermr (approved):

scripts/delete_staged_recipes_token.py
```diff
@@ -43,10 +45,12 @@ def feedstock_token_exists(organization, name):
                 shell=True,
             )
 
+            msg = with_action_url(
+                "'[ci skip] [skip ci] [cf admin skip] ***NO_CI*** "
```
beckermr, line 49:
```suggestion
                "[ci skip] [skip ci] [cf admin skip] ***NO_CI*** "
```

scripts/test_cfep13_copy.py
```diff
@@ -78,7 +78,7 @@ def _clone_and_remove(repo, file_to_remove):
                     _run_git_command(
                         "commit",
                         "-m",
-                        "'removed %s for testing'" % file_to_remove,
+                        with_action_url(f"'removed {file_to_remove} for testing'"),
```
beckermr, line 81:
```suggestion
                        "'%s'" % with_action_url(f"removed {file_to_remove} for testing"),
```

scripts/test_cfep13_endpoints.py
```diff
@@ -67,7 +67,7 @@ def _clone_and_remove(repo, file_to_remove):
                     _run_git_command(
                         "commit",
                         "-m",
-                        "'removed %s for testing'" % file_to_remove,
+                        with_action_url(f"'removed {file_to_remove} for testing'"),
```
beckermr, line 70:
```suggestion
                        "'%s'" % with_action_url(f"removed {file_to_remove} for testing"),
```

jakirkham:
Think we need to add autoflake to make that work

That or add black or ruff or something

scripts/test_cfep13_copy.py
```diff
@@ -78,7 +78,7 @@ def _clone_and_remove(repo, file_to_remove):
                     _run_git_command(
                         "commit",
                         "-m",
-                        "'removed %s for testing'" % file_to_remove,
+                        "'%s'" % with_action_url(f"removed {file_to_remove} for testing"),
```
jakirkham, line 81:
```suggestion
                        with_action_url(f"removed {file_to_remove} for testing"),
```

scripts/test_cfep13_endpoints.py
```diff
@@ -67,7 +67,7 @@ def _clone_and_remove(repo, file_to_remove):
                     _run_git_command(
                         "commit",
                         "-m",
-                        "'removed %s for testing'" % file_to_remove,
+                        "'%s'" % with_action_url(f"removed {file_to_remove} for testing"),
```
jakirkham, line 70:
```suggestion
                        with_action_url(f"removed {file_to_remove} for testing"),
```

jakirkham (commented):
Thanks Jaime! 🙏

These are already `str`s right? Or at least that it is what the function typing says. Think we can simplify these (can thus fix the lint)

Or do we need the extra quotes. If so, maybe we can have a `commit_msg` variable constructed before hand with the message format and then use that in the `git commit` calls

scripts/test_cfep13_copy.py
beckermr: It needs to be quoted in single quotes so this change is not correct.

scripts/test_cfep13_endpoints.py
beckermr: Same comment here.

scripts/test_cfep13_copy.py
jakirkham: Is there some more context on why?

Also please see review OP suggestion for other options already outlined (should this be the case)

beckermr: I don't understand. It is executing a command at the cli that looks like thjs 

git commit -m 'message with spaces' 


You have to have quotes. I am saying to use single quotes inside double quotes.

jakirkham: Am not finding that to be the case

Made a small tweak in a feedstock (used `setuptools` locally, but that shouldn't matter)

```zsh
% git add -p
diff --git a/conda-forge.yml b/conda-forge.yml
index 442becc..14aeb90 100644
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -1,3 +1,4 @@
+
 bot:
   inspection: update-grayskull
   abi_migration_branches:
(1/1) Stage this hunk [y,n,q,a,d,e,?]? y
```

```python
In [1]: import subprocess

In [2]: subprocess.run(["git", "commit", "-m", "Add blank space"])
[detached HEAD 5453917] Add blank space
 1 file changed, 1 insertion(+)
Out[2]: CompletedProcess(args=['git', 'commit', '-m', 'Add blank space'], returncode=0)
```

```zsh
% git log --pretty=oneline -n1
545391757af4272b68a851a657c21d87c5b96a73 (HEAD) Add blank space
```

Please let me know if I'm missing something

scripts/test_cfep13_endpoints.py
jakirkham: Let's discuss in one thread ( https://github.com/conda-forge/conda-forge-webservices/pull/609#discussion_r1550651937 )

Then we can handle both the same way

scripts/test_cfep13_copy.py
beckermr: Please see the actual call to subprocess.run


def _run_git_command(*args):
    subprocess.run(
        " ".join(["git"] + list(args)),
        check=True,
        shell=True,
    )


We are building the shell line explicitly, not using the list of arguments which would quite for us.

beckermr: > What was the motivation for not using a list? What doesn't work with a list?

No motivation one way or the other.

You said above the quotes were not needed according to your tests in the code snippets. 

I was explaining why the quotes are needed in this case.

jaimergp:
Thanks Matt and John! 🙏 

Having peeked at this code now, is there any reason to use `shell=True` beyond the convenience of writing things without a list of subcommands? We wouldn't have to think about quoting in that case.

beckermr:
I have a long habit of always using shell=True. That is a personal quirk. Idk when or why one would use this or not use it.

jakirkham:
Personally internalized [the security warning regarding `shell=True`]( https://docs.python.org/3/library/subprocess.html#security-considerations ) and so try never to use it

Would suggest we try eliminating `shell=True` wherever possible

PR #626, opened 2024-06-27
clean up code a little bit

wolfv:
Remove some function duplication and don't use `os.path.sep` for URLs.


beckermr (approved):
Thank you and LGTM!

PR #631, opened 2024-07-16
feat: handle recipe.yaml when linting

wolfv:
Also a bit of a pathlib refactor.

### DIFF
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -4,7 +4,7 @@ import time
 from tempfile import TemporaryDirectory
 import logging
 from pathlib import Path
-from typing import TypedDict, Optional
+from typing import TypedDict, Optional, List
 
 from git import GitCommandError, Repo
 import github
@@ -21,7 +21,7 @@ class LintInfo(TypedDict):
     sha: str
 
 
-def find_recipes(path: Path) -> [Path]:
+def find_recipes(path: Path) -> List[Path]:
     """
     Returns all `meta.yaml` and `recipe.yaml` files in the given path.
     """
@@ -44,7 +44,7 @@ def compute_lint_message(
         time.sleep(1.0)
         pull_request = remote_repo.get_pull(pr_id)
         if pull_request.state != "open":
-            return {}
+            return None
         mergeable = pull_request.mergeable
 
     tmp_dir = TemporaryDirectory(suffix="_recipe")
@@ -99,9 +99,7 @@ def compute_lint_message(
                 """)  # noqa
             status = "merge_conflict"
 
-            lint_info: LintInfo = {"message": message, "status": status, "sha": sha}
-
-            return lint_info
+            return {"message": message, "status": status, "sha": sha}
 
         # Collect recipes from base that should be ignored.
         base_recipes = []
@@ -234,11 +232,9 @@ def compute_lint_message(
 
     pull_request = remote_repo.get_pull(pr_id)
     if pull_request.state == "open":
-        lint_info: LintInfo = {"message": message, "status": status, "sha": sha}
+        return {"message": message, "status": status, "sha": sha}
     else:
-        lint_info = None
-
-    return lint_info
+        return None
 
 
 def comment_on_pr(
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -208,7 +208,7 @@ def compute_lint_message(
         commit_msg = repo.commit(sha).message
         should_skip = any([msg in commit_msg for msg in skip_msgs])
         if should_skip:
-            return {}
+            return None
 
         # Raise an error if the PR is not mergeable.
         if not mergeable:
--- a/conda_forge_webservices/linting.py
+++ b/conda_forge_webservices/linting.py
@@ -35,72 +35,64 @@ def lint_all_recipes(all_recipe_dir: Path, base_recipes: List[Path]) -> (str, st
     """
     Lint all recipes in the given directory.
     """
-    try:
-        recipes = find_recipes(all_recipe_dir)
-        all_pass = True
-        messages = []
-        hints = []
+    recipes = find_recipes(all_recipe_dir)
+    all_pass = True
+    messages = []
+    hints = []
 
-        # Exclude some things from our list of recipes.
-        # Sort the recipes for consistent linting order (which glob doesn't give us).
-        pr_recipes = sorted(set(recipes) - set(base_recipes))
+    # Exclude some things from our list of recipes.
+    # Sort the recipes for consistent linting order (which glob doesn't give us).
+    pr_recipes = sorted(set(recipes) - set(base_recipes))
 
-        rel_pr_recipes = []
-        for recipe in pr_recipes:
-            recipe_dir = recipe.parent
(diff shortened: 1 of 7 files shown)


nichmor (approved):

beckermr (changes requested):
We should use a status of failed for recipes we cannot yet lint?

beckermr:
Also we should add a test.

beckermr:
Yes look in the code to see the test suite.

beckermr:
I'm all for ruff here, but it'd be better to bring that in in a separate pr instead the piecemeal changes.

wolfv:
sure, i only changed the `linting.py` for now and didn't push the precommit changes.

conda_forge_webservices/linting.py
```diff
@@ -176,12 +94,14 @@ def compute_lint_message(
...
                     )
                 )
-
-    finally:
-        # Remove the environment variable if it was set in this function
-        os.environ.pop("STAGED_RECIPES_PR_NUMBER", None)
-
-        tmp_dir.cleanup()
+    except Exception as e:
+        LOGGER.error("Error while linting recipes: %s", repr(e))
```
beckermr, line 98:
```suggestion
```

Repeated code here.

wolfv:
@beckermr i added two test (one snapshot test, one normal test). Happy to remove the snapshot test. 

I also refactored the code a bit so that we can test the linting message generation more easily without requiring to open a PR on this repository.

beckermr:
Thanks @wolfv! I will take a look today.

conda_forge_webservices/tests/linting/test_compute_lint_message.py
```diff
@@ -1,197 +1,222 @@
...
+    assert lint is None
+
+
+def test_new_recipe(tmp_path, snapshot):
+    recipe_file = tmp_path / "recipe" / "recipe.yaml"
+    recipe_file.parent.mkdir(parents=True)
+    shutil.copy(data_folder() / "recipe.yaml", recipe_file)
+
+    message = lint_all_recipes(Path(tmp_path), [])
+    assert message == snapshot(name="rattler-build-good-lint")
```
beckermr, line 222: What is the convention for how to match up serialized test results to the tests? The name you specified here has nothing to do with the name of the test, and so makes it hard to match up.

beckermr (changes requested):
I have one question about the snapshot tests. TBH I don't like it right now since it is very opaque as to what is being testing and where to find the actual expected value.

conda_forge_webservices/tests/linting/test_compute_lint_message.py
wolfv: Hmm, it's `test_new_recipe[rattler-build-good-lint]`, ie. the name of the function + the `name` property.

```
test_new_recipe[rattler-build-good-lint]
```

How should we proceed?

beckermr: So what is the name argument for then?

I don't know what to expect here and am trying to keep an open mind.

wolfv: It's unnecessary. Without the name argument I think it will be something like `test_new_recipe[1]`.

beckermr: Ah got it. What indicates to future developers that the snapshot here should correspond to a passing recipe? That is a constraint this test is trying to enforce and I don't see where it is located.

wolfv: Should we try `inline-snapshot`? https://github.com/15r10nk/inline-snapshot

beckermr: Yep. That looks better to me personally.

More annoying questions from me. Sorry. 

What then is the purpose of this test? Is it to make sure the linter runs? To make sure it passes on this recipe? Something else?

As done now the test becomes brittle since the linter output could change, but depending on the purpose of the test, that could be just fine.

wolfv: I just copied the style of many other tests in this file. Is that not what you wanted?

beckermr: Right OK. Many of the existing tests are not precise enough and ambiguous. 

Here is what I mean. I've put examples of different possible tests below, each of which tests different things.

## test things run at all

```python
def test_linter_runs_at_all():
    # no asserts here, test fails if code fails
    call_inter(...)
```

## test things run and produce an output

```python
def test_linter_runs_with_some_output():
    val = call_inter(...)
    assert val is not None
```

## test things run and produce a pass for linting

```python
def test_linter_runs_and_passes():
    # the assert here is robust to changes in linter hints
    val = call_inter(...)
    assert "excellent condition" in val["message"]
```

## test things run and produce a pass for linting w/ hints

```python
def test_linter_runs_and_passes():
    val = call_inter(...)
    assert "excellent condition" in val["message"]
    assert "suggestions for making" in val["message"]
```

## test a specific message is produced in an edge case

```python
def test_linter_with_conflict():
    # this value is built by this code base, not smithy,
    # and so hard coding it here robust to changes in smithy
    expected_value = ...  
    val = call_inter(...)
    assert val["message"] == expected_value
```

## test a specific message is produced in an edge case

```python
def test_linter_runs_with_exact_message():
    # this value is built by a combination of this code base 
    # and smithy. Having it hard coded here is not robust to changes 
    # in smithy.
    expected_value = ...
    val = call_inter(...)
    assert val["message"] == expected_value
```

I don't see how to use snapshotting tools to write tests that are specific and robust.  Snapshotting certainly covers the last case nicely since it automates the updates due to possible changes in smithy. Inline snapshotting is even better since the diff will be in a single file.

wolfv: Hm, I am not sure if I get it. To me, snapshotting is definitely precise (the same as what the current code is doing). 

What is nice is that updating the snapshots is much easier (or would be, if we could run the tests locally).

If you would change some code, you would just run `pytest --update-snapshots`, and compare the diffs of the snapshots to make sure you're happy. It also makes it much easier to add lots of (precise) tests by just adding `message = snapshot()` and then the snapshot is inserted automatically (inline, once we switch to that).

beckermr: Ok I think we've reached the philosophical crux of the issue. 

The set of tests I coded above in this comment thread each test one specific (possibly additional) thing about the linter. They also "telescope" in the sense that depending on which ones fail and which ones pass, you can get a sense of what might be broken. This feature makes the test suite useful. 

Otoh if one just has snapshot tests, you loose some of the debugging ability because either it passes by producing exactly the same output, or it doesn't. One would have to examine the outputs of all of the tests, figure out which features are common or different for ones that do or don't pass, and then infer what might have happened. 

So I'm fine with inline snapshotting for tests that examine the entire output of the linter message. However, I don't want to convert tests that have specific asserts beyond pure '==' because it hampers debugging in many cases.

beckermr:
pre-commit.ci autofix

wolfv:
added noqa.

beckermr (approved):

PR #641, opened 2024-07-22
remove try import in preparation for mypy

wolfv:
Checklist
* [x] CI is passing

### DIFF
--- a/conda_forge_webservices/tests/test_commands.py
+++ b/conda_forge_webservices/tests/test_commands.py
@@ -1,10 +1,6 @@
 import os
 import unittest
-
-try:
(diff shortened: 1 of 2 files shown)


beckermr:
Why can't mypy handle try import patterns?

wolfv:
It might, but do we still need this? This seems deprecated since Python 3.3 which is long EOL.

wolfv:
Even `ruff` autocorrects this to the import from `unittest`.

beckermr:
Oh sorry. Yes this pr is fine. I was wondering why this was mypy related.

PR #642, opened 2024-07-22
Only automated fixes to update the code style using ruff

wolfv:
This did also check a section that was removed in #641 - so I'd regenerate that then.


wolfv:
This was done using `ruff check --fix --unsafe-fixes`.
No other changes (except for adding `# noqa` in two places.

PR #643, opened 2024-07-22
feat: modify the add user command to work for `recipe.yaml` as well

tdejager:
This adds the `add-user` functionality for `rattler-build` `recipe.yaml` as well.

Checklist
* [x] CI is passing

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -740,9 +740,14 @@ def add_user(repo, user):
     # adjust those easily enough.
     # Those happen to also be the only locations where adding a user really matters.
 
+    curr_users = None
+    # Initialize data for the different recipe types
     recipe_path = os.path.join(repo.working_dir, "recipe", "meta.yaml")
-    co_path = os.path.join(repo.working_dir, ".github", "CODEOWNERS")
+    rattler_build_recipe_path = os.path.join(repo.working_dir, "recipe", "recipe.yaml")
+    rattler_build_data = None
+
     yaml = YAML(typ="safe")
+    # This is an conda-build recipe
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
We don't need this pr. The code as is works with both formats since they both have the same extra section.

tdejager:
Ok you are right! Closing this and #664

conda_forge_webservices/commands.py
```diff
@@ -740,9 +740,14 @@ def add_user(repo, user):
     # adjust those easily enough.
     # Those happen to also be the only locations where adding a user really matters.
 
+    curr_users = None
+    # Initialize data for the different recipe types
     recipe_path = os.path.join(repo.working_dir, "recipe", "meta.yaml")
```
isuruf, line 745: All we want is to fix this variable. Everything else works fine.

beckermr: good point @isuruf ! We likely need just this kind of fix in a few other places too.

tdejager: Well, there could be a case that someone is using inline array, which would technically be allowed right?

e.g: ["user1", "user2"]

But that would fail for the current code as well I think?

beckermr: Nope. We strip everything from the recipe except the extra section and then parse it via yaml to catch this case.

tdejager: Alright!

Wat was is meant by @isuruf with 'fix' can I take this into account in the other PR?

beckermr: Isuru just meant fixing the path to the recipe IIUIC.

PR #644, opened 2024-07-22
feat: make noarch available for rattler-build recipe as well

tdejager:
Makes the `make_noarch` command available for `recipe.yaml`. as well.

Checklist
* [x] CI is passing

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -948,30 +948,72 @@ def update_version(full_name, pr_num, input_ver):
 
     return not repo.create_repository_dispatch(
         "version_update",
-        client_payload={"pr": pr_num, "input_version": input_ver or 'null'},
+        client_payload={"pr": pr_num, "input_version": input_ver or "null"},
     )
 
 
-def make_noarch(repo):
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/commands.py
```diff
@@ -948,30 +948,72 @@ def update_version(full_name, pr_num, input_ver):
...
+                    spaces = len(line) - len(line.lstrip())
+                    line = "{}noarch: python\n{}".format(" " * spaces, line)
+                build_line = False
+                if line.rstrip() == "build:":
+                    build_line = True
+                fh.write(line)
+        return meta_location
+    except Exception:
+        LOGGER.error("Error when trying to add noarch: python to meta.yaml")
+        return None
```
isuruf, line 973: What kind of an exception do you expect here?

tdejager: Maybe a that the file cannot be opened? Shouldn't happen because of the check, but still maybe if some code moves around? If you rather take it out I'm fine doing that :)

beckermr (changes requested):
Again we do not need this pr since the existing code works with both formats. Extra code we don't need is more maintenance.

PR #645, opened 2024-07-22
feat: change command recipe path

tdejager:
This supersedes #644 and #643.

Only modifies the recipe path, so that the commands are also applied to rattler-build recipe.yaml.

cc @isuruf @beckermr 

Checklist
* [x] CI is passing

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -966,7 +966,7 @@ def update_version(full_name, pr_num, input_ver):
 def make_noarch(repo):
     meta_yaml = _determine_recipe_path(repo)
     if meta_yaml is None:
-        return None
+        return False
     with open(meta_yaml, 'r') as fh:
         lines = [line for line in fh]
     with open(meta_yaml, 'w') as fh:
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
Please make this PR without all of the unassociated style changes to the single vs double quotes. Maybe black was applied? I cannot read these diffs and figure out what is a style versus a functional changed.

I am ok with black or ruff-format or w/e here, but we need to have one thing per PR.

tdejager:
Yeah sorry, somehow vscode was applying these automatically, will revert those changes tomorrow.

tdejager:
Changed it to just contain the changes. Two notes/questions.

I'm returning None from the `add_user` as that is checked, and False from the `make_noarch`. Don't know if you guys agree with that, we can also throw an Exception, ignore, or add a `None` check to `make_noarch` places.

beckermr (approved):

beckermr:
Ideally the admin would make a comment that the recipe was missing. We can do that later if we want.

PR #647, opened 2024-07-23
chore: make ruff fix files

beckermr:
Checklist
* [ ] CI is passing


beckermr:
cc @wolfv

beckermr:
We were supposed to be running ruff already and I think the lack of `--fix` cause pre-commit to pass when it should not have maybe? It is confusing to me and I may be wrong.

wolfv:
I think before it would've still failed if there was lint :) But now it's gonna fix it, too 🚀

PR #665, opened 2024-09-09
feat: use pending status for linting

beckermr:
Checklist
* [ ] CI is passing


beckermr:
OK @conda-forge/core. This PR is ready. It sets the pending status properly when the linter is running instead of just setting the result. This should help folks get more insight into what is going on.

xhochy (approved):
You could alternatively also mock the method instead of adding the additional flag. I have no hard opinion on it, so fine to merge (for me).

PR #709, opened 2024-10-01
feat: squash the docker image

### LINKED ISSUE #707: Squash Docker image?
Noticed this image has quite a few layers. This can add to the time to pull the image. Noting this since each bot operation pulls the image

Am wondering if we can do something to consolidate the image layers. For example in `docker-images` we enabled squash ( https://github.com/conda-forge/docker-images/pull/99 ), which helped address a similar problem with those images. Am wondering if we could apply a similar approach here

beckermr:
### Description

This PR enables docker image squashing. 

fixes #707

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -114,25 +114,48 @@ jobs:
           version=$(python -c "import conda_forge_webservices; print(conda_forge_webservices.__version__.replace('+', '.'))")
           echo "version=${version}" >> "$GITHUB_OUTPUT"
 
-      - name: set up docker buildx
-        if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'conda-forge/conda-forge-webservices'
-        uses: docker/setup-buildx-action@988b5a0280414f521da01fcc63a27aeeb4b104db # v3
+      - name: enable experimental docker features (needed for squash)
+        run: |
+          sudo cp docker_daemon_config.json /etc/docker/daemon.json
+          sudo service docker restart
 
-      - name: login to docker hub
-        if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'conda-forge/conda-forge-webservices'
-        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3
-        with:
-          username: condaforgebot
-          password: ${{ secrets.CF_BOT_DH_PASSWORD }}
+      - name: docker info
+        run: |
+          docker info
 
-      - name: build and push docker image
-        if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'conda-forge/conda-forge-webservices'
-        uses: docker/build-push-action@4f58ea79222b3b9dc2c8bbdd6debcef730109a75 # v5
-        with:
-          context: .
-          file: Dockerfile_wda
-          push: true
-          tags: condaforge/webservices-dispatch-action:${{ steps.install-code.outputs.version }}
+      - name: build and push
+        run: |
+          image_tag=condaforge/webservices-dispatch-action:${{ steps.install-code.outputs.version }}
+
+          docker build \
+            -t ${image_tag} \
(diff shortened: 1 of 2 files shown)


beckermr:
None of this seems to help. I think we should adopt the approach in the automerge action which generates an image that is ~7x smaller than this one.

Dockerfile_wda
```diff
@@ -1,23 +1,43 @@
...
+COPY conda-lock.yml /tmp/conda-lock.yml
+
+RUN echo "**** install base env ****" && \
+    micromamba create --yes --quiet --name cf-feedstock-ops --file /tmp/conda-lock.yml
+RUN echo "**** cleanup ****" && \
+    micromamba clean --all --force-pkgs-dirs --yes && \
+    find "${MAMBA_ROOT_PREFIX}" -follow -type f \( -iname '*.a' -o -iname '*.pyc' -o -iname '*.js.map' \) -delete
+RUN echo "**** finalize ****" && \
+    mkdir -p "${MAMBA_ROOT_PREFIX}/locks" && \
+    chmod 777 "${MAMBA_ROOT_PREFIX}/locks"
```
jakirkham, line 18: Think we can use [heredocs]( https://docs.docker.com/reference/dockerfile/#here-documents ) (introduced in [this blogpost]( https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/ )) to combine these into a single `RUN` command/layer while also making the syntax more manageable

```suggestion
RUN <<EOF
echo "**** install base env ****"
micromamba create --yes --quiet --name cf-feedstock-ops --file /tmp/conda-lock.yml
echo "**** cleanup ****"
micromamba clean --all --force-pkgs-dirs --yes
find "${MAMBA_ROOT_PREFIX}" -follow -type f \( -iname '*.a' -o -iname '*.pyc' -o -iname '*.js.map' \) -delete
echo "**** finalize ****"
mkdir -p "${MAMBA_ROOT_PREFIX}/locks"
chmod 777 "${MAMBA_ROOT_PREFIX}/locks"
EOF
```

Dockerfile_wda
```diff
@@ -1,23 +1,43 @@
...
+
+FROM frolvlad/alpine-glibc:alpine-3.16_glibc-2.34
+LABEL maintainer="conda-forge <conda-forge@googlegroups.com>"
+
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
 ENV TMPDIR=/tmp
 ENV CF_FEEDSTOCK_OPS_DIR=/opt/cf-feedstock-ops
 ENV CF_FEEDSTOCK_OPS_ENV=cf-feedstock-ops
```
jakirkham, line 27: Think we can also join these together like so. See something similar in [this doc]( https://docs.docker.com/reference/dockerfile/#env )

```suggestion
ENV LANG=en_US.UTF-8 \
    LANGUAGE=en_US.UTF-8 \
    TMPDIR=/tmp \
    CF_FEEDSTOCK_OPS_DIR=/opt/cf-feedstock-ops \
    CF_FEEDSTOCK_OPS_ENV=cf-feedstock-ops
```

Dockerfile_wda
beckermr: done!

Dockerfile_wda
beckermr: done too!

environment.yml
```diff
@@ -14,6 +15,7 @@ dependencies:
   - conda-forge-feedstock-ops >=0.7.0
   - conda-smithy >=3.34.1  # this pin is for jinja2 sandboxes
   - cryptography >=39
+  - docker-squash
```
beckermr, line 18:
```suggestion
```

environment.yml
```diff
@@ -3,6 +3,7 @@ channels:
   - conda-forge
 platforms:
   - linux-64
+  - linux-aarch64
```
beckermr, line 6:
```suggestion
```

beckermr:
pre-commit.ci autofix

Issue #743, opened 2024-10-24
permissions error on executable file after rerender when file is new


beckermr:
I think we may have some residual permissions errors for executable files during rerenders: https://github.com/conda-forge/stream-unzip-feedstock/pull/27/commits/b6b74324d43b0e933a30f22e0347d77ab45b2ae2

cc @hmaarrfk

hmaarrfk:
Just hit this
https://github.com/conda-forge/qt-gtk-platformtheme-feedstock/pull/25
https://github.com/conda-forge/qt-wayland-feedstock/pull/25
```
 mode change 100644 => 100755 .azure-pipelines/azure-pipelines-linux.yml
 mode change 100644 => 100755 .scripts/build_steps.sh
 mode change 100644 => 100755 .scripts/run_docker_build.sh
 mode change 100644 => 100755 build-locally.py
```

i guess the release didn't capture the fix.

hmaarrfk:
Even more beautifully https://github.com/conda-forge/qt-wayland-feedstock/pull/25 actually changed the permission from executable to not, so i had to rerender.

beckermr:
Ouch. Do you have a link to the gha job?

beckermr:
OK. #748 appears to fix things: https://github.com/conda-forge/conda-forge-webservices/actions/runs/11543400624/job/32127780380

I think github removes all execute file permissions when it downloads artifacts or something. Instead I am passing around the actual git patch which seems to be robust.

PR #747, opened 2024-10-27
fix: turn off permissions changes for now

beckermr:
### Description

This PR turns off the permissions sync to see if that helps for now.


beckermr:
xref #743

PR #768, opened 2024-11-09
feat: set target url

beckermr:
### Description

This PR sets the target url for the pending linter to see if we can get it to appear in the PR checks.


conda_forge_webservices/github_actions_integration/linting.py
```diff
@@ -158,13 +158,17 @@ def build_and_make_lint_comment(gh, repo, pr_id, lints, hints):
...
-            """
+        mixed = (
+            good
+            + "\n"
+            + dedent_with_escaped_continue(
+                """
             I do have some suggestions for making it better though...
 
             {}
             """
```
beckermr, line 169:
```suggestion
                """
                I do have some suggestions for making it better though...

                {}
                """
```

PR #902, opened 2025-03-11
Explain on the recommended way to be added as maintainer

hmaarrfk:
xref: https://github.com/conda-forge/xarray-dataclasses-feedstock/pull/7#issuecomment-2712288200
xref: https://conda-forge.org/docs/maintainer/infrastructure/#conda-forge-admin-please-add-user-username

### Description

I feel like this would help establish the policy conda-forge wide.

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -605,6 +605,13 @@ def issue_comment(org_name, repo_name, issue_num, title, comment, comment_id=Non
                             pr_title = f"[ci skip] adding user @{user}"
                             to_close = ADD_USER.search(title)
                             extra_msg = (
+                                "\n\nWPlease note that we no longer recommend this as the "  # noqa
+                                "as the way to add new contributors to a feedstock. "  # noqa
+                                "To aid with vetting new end users and maintainers, we "  # noqa
+                                "recommend that you make a PR that fixes the recipe in a "  # noqa
+                                "meaningful way and then ask the existing maintainers to "  # noqa
+                                "review it. Near the end of the review process, you may ask " # noqa
+                                "to be added as a maintainer. "  # noqa
                                 "\n\nMerge this PR to add the user. Please do not rerender "  # noqa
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/commands.py
```diff
@@ -605,6 +605,13 @@ def issue_comment(org_name, repo_name, issue_num, title, comment, comment_id=Non
                             pr_title = f"[ci skip] adding user @{user}"
                             to_close = ADD_USER.search(title)
                             extra_msg = (
+                                "\n\nWPlease note that we no longer recommend this as the "  # noqa
+                                "as the way to add new contributors to a feedstock. "  # noqa
+                                "To aid with vetting new end users and maintainers, we "  # noqa
+                                "recommend that you make a PR that fixes the recipe in a "  # noqa
+                                "meaningful way and then ask the existing maintainers to "  # noqa
+                                "review it. Near the end of the review process, you may ask " # noqa
+                                "to be added as a maintainer. "  # noqa
```
beckermr, line 614: I think we're confusing two issues, namely:

1. The expected etiquette about how to be added to a feedstock
2. The technical details of how to actually do that task.

The bot command concerns the second item whereas the first item is something different. 

We do actually recommend the bot command as the best way to add folks to feedstocks in the absence of another PR. Otherwise folks forget things like ci skip and possibly even bump build numbers pushing new packages that don't need to be pushed.

Maybe instead we can make a PR to the docs with the expectations for when and how people ask to be added to a feedstock? 

Then we can adjust the text here to link to the docs and have a more neutral message like "Hey! Did you ready the docs on how become a maintainer of a feedstocks? 
Make sure to do that."

hmaarrfk: i guess that works. but i'm seeing this alot and I would like to stop the bleed somehow. i'll see what i can do to adapt this to the docs. 

will have to circle back eventually. if somebody wants to jump on it that is great too.

PR #915, opened 2025-03-20
fix: use a random number to ensure cache is busted on heroku

beckermr:
### Description

cc @jaimergp

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,8 @@ ENV PYTHONDONTWRITEBYTECODE=1
 
 # bust the docker cache so that we always rerun the installs below
 # ADD https://loripsum.net/api /opt/docker/etc/gibberish
-ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish
+# ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish
+ADD http://www.randomnumberapi.com/api/v1.0/randomnumber /opt/docker/etc/gibberish
 
 # Install conda
 COPY conda-lock.yml /
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ ENV PYTHONDONTWRITEBYTECODE=1
 # bust the docker cache so that we always rerun the installs below
 # ADD https://loripsum.net/api /opt/docker/etc/gibberish
 # ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish
-ADD https://www.randomnumberapi.com/api/v1.0/randomnumber?min=1&max=10000000&count=100 /opt/docker/etc/gibberish
+ADD https://www.random.org/cgi-bin/randbyte?nbytes=4096&format=h /opt/docker/etc/gibberish
 
 # Install conda
 COPY conda-lock.yml /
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -40,7 +40,8 @@ ENV PYTHONDONTWRITEBYTECODE=1
 
 # bust the docker cache so that we always rerun the installs below
 # ADD https://loripsum.net/api /opt/docker/etc/gibberish
-ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish
+# ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish
+ADD http://www.randomnumberapi.com/api/v1.0/randomnumber /opt/docker/etc/gibberish
```
jaimergp, line 44:
```suggestion
ADD https://www.randomnumberapi.com/api/v1.0/randomnumber?min=1&max=100000&count=5 /opt/docker/etc/gibberish
```

Dockerfile_wda
```diff
@@ -5,7 +5,8 @@ USER root
 
 # make sure the install below is not cached by docker
 # ADD https://loripsum.net/api /opt/docker/etc/gibberish-to-bust-docker-image-cache
-ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish-to-bust-docker-image-cache
+# ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish-to-bust-docker-image-cache
+ADD http://www.randomnumberapi.com/api/v1.0/randomnumber /opt/docker/etc/gibberish-to-bust-docker-image-cache
```
jaimergp, line 9:
```suggestion
ADD https://www.randomnumberapi.com/api/v1.0/randomnumber?min=1&max=100000&count=5 /opt/docker/etc/gibberish-to-bust-docker-image-cache
```

jaimergp:
Let's change the params a bit because the default only returns 1 number between 1 and 100 and it's not enough entropy (got [9] two times in a row after just a few attempts).

Dockerfile
beckermr: They don't support https. I guess that means we need a different url then.

Dockerfile
```diff
@@ -40,7 +40,8 @@ ENV PYTHONDONTWRITEBYTECODE=1
 
 # bust the docker cache so that we always rerun the installs below
 # ADD https://loripsum.net/api /opt/docker/etc/gibberish
-ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish
+# ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish
+ADD https://www.randomnumberapi.com/api/v1.0/randomnumber?min=1&max=10000000&count=100 /opt/docker/etc/gibberish
```
beckermr, line 44:
```suggestion
ADD https://www.random.org/cgi-bin/randbyte?nbytes=4096&format=h /opt/docker/etc/gibberish
```

Dockerfile_wda
```diff
@@ -5,7 +5,8 @@ USER root
 
 # make sure the install below is not cached by docker
 # ADD https://loripsum.net/api /opt/docker/etc/gibberish-to-bust-docker-image-cache
-ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish-to-bust-docker-image-cache
+# ADD https://worldtimeapi.org/api/timezone/Etc/UTC /opt/docker/etc/gibberish-to-bust-docker-image-cache
+ADD https://www.randomnumberapi.com/api/v1.0/randomnumber?min=1&max=10000000&count=100 /opt/docker/etc/gibberish-to-bust-docker-image-cache
```
beckermr, line 9:
```suggestion
ADD https://www.random.org/cgi-bin/randbyte?nbytes=4096&format=h /opt/docker/etc/gibberish-to-bust-docker-image-cache
```

PR #957, opened 2025-04-11
perf: reduce memory usage again

beckermr:
### Description

I am still seeing some out of memory issues.

### DIFF
--- a/conda_forge_webservices/feedstock_outputs.py
+++ b/conda_forge_webservices/feedstock_outputs.py
@@ -76,10 +76,10 @@ def is_valid_feedstock_token(user, project, feedstock_token, provider=None):
     return False
 
 
-def _get_ac_api_with_timeout(token, timeout=10):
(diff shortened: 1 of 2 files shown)


jakirkham:
Ah is OOM what is causing the validation issues?

beckermr:
No it is an actual failure of the anaconda.org api for copying packages. It only happens on the pinnings package right now. The OOM issue is causing heroku to reboot the server every few hours which is disruptive and not ideal. This PR also increased the timeout on api requests to anaconda.org to see if we can get the call to pass.

jakirkham:
Ah ok. Does the Anaconda team know? Is it worth raising [an infrastructure issue]( https://github.com/conda/infrastructure/issues )?

Gotcha. Perhaps it is worth doing another trim of the Docker image. Though admittedly we did a pretty good job last time. Maybe there are some other tricks we can try?

PR #1091, opened 2025-09-05
perf: try commands in a process pool outside of copy process pool

beckermr:
### Description

At this point, all of the commands that do significant work delegate jobs to GHA. So I think we can try to run commands in a separate process pool so they are responsive. Currently commands are run in the same process pool as uploads and this causes issues when lots of uploads are happening.

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -76,6 +76,9 @@ def _worker_pool(kind):
                 # needed for mocks in testing
                 COMMAND_POOL = ThreadPoolExecutor(max_workers=2)
             else:
+                # we only use one process here to conserve memory.
+                # we have to use a process because the commands
+                # run git operations which are not thread safe.
                 COMMAND_POOL = ProcessPoolExecutor(max_workers=1)
         return COMMAND_POOL
     elif kind == "upload":
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -70,15 +70,15 @@ def _worker_pool(kind):
     global UPLOAD_POOL
     global COPYLOCK
 
-    # if kind == "command":
-    #     if COMMAND_POOL is None:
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/webapp.py
```diff
@@ -70,15 +70,15 @@ def _worker_pool(kind):
...
-    #             COMMAND_POOL = ProcessPoolExecutor(max_workers=2)
-    #     return COMMAND_POOL
-    if kind == "upload" or kind == "command":
+    if kind == "command":
+        if COMMAND_POOL is None:
+            if "PYTEST_CURRENT_TEST" in os.environ:
+                # needed for mocks in testing
+                COMMAND_POOL = ThreadPoolExecutor(max_workers=2)
+            else:
+                COMMAND_POOL = ThreadPoolExecutor(max_workers=2)
```
beckermr, line 79:
```suggestion
                COMMAND_POOL = ProcessPoolExecutor(max_workers=1)
```

conda_forge_webservices/webapp.py
```diff
@@ -70,15 +70,15 @@ def _worker_pool(kind):
...
-    #             COMMAND_POOL = ProcessPoolExecutor(max_workers=2)
-    #     return COMMAND_POOL
-    if kind == "upload" or kind == "command":
+    if kind == "command":
+        if COMMAND_POOL is None:
+            if "PYTEST_CURRENT_TEST" in os.environ:
+                # needed for mocks in testing
+                COMMAND_POOL = ThreadPoolExecutor(max_workers=2)
+            else:
+                COMMAND_POOL = ProcessPoolExecutor(max_workers=1)
```
beckermr, line 79:
```suggestion
                # we only use one process here to conserve memory.
                # we have to use a process because the commands
                # run git operations which are not thread safe.
                COMMAND_POOL = ProcessPoolExecutor(max_workers=1)
```

PR #1097, opened 2025-09-19
feat: rebuild docker images when webservices is relocked

beckermr:
### Description

@jaimergp would this work? It should help us rebuild less often maybe?

### DIFF
--- a/.github/workflows/clean-and-update.yml
+++ b/.github/workflows/clean-and-update.yml
@@ -16,7 +16,8 @@ jobs:
           token: ${{ secrets.CF_ADMIN_GITHUB_TOKEN }}
 
       - name: relock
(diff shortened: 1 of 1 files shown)


.github/workflows/clean-and-update.yml
```diff
@@ -28,13 +29,39 @@ jobs:
...
+            conda-forge-ci-setup
             conda-forge-tick
             conda-forge-feedstock-ops
             conda-recipe-manager
             conda-souschef
             rattler-build
             rattler-build-conda-compat
 
+      - name: generate token
+        id: generate_token
```
beckermr, line 41:
```suggestion
        id: generate_token
        if: steps.relock.outputs.relocked == 'true'
```

PR #1121, opened 2025-10-06
fix: ensure we check authors and committers for automerge

### LINKED ISSUE #1120: Also check committers for automerge
As correctly analyzed by @shermansiu in https://github.com/conda-forge/xformers-feedstock/pull/58:
https://github.com/conda-forge/conda-forge-webservices/blob/9f283a7b21ed04213ae8141bda5a752a695324a9/conda_forge_webservices/github_actions_integration/automerge.py#L384-L386

does not distinguish between the _authors_ and _committers_ of a commit, which do not necessarily coincide. For example, authorship and committership will diverge when rebasing a PR onto a main branch that has moved on in the meantime (though there are other ways of course).

So we should check that both authors and committers are in `ALLOWED_USERS ` for automerge to pass.

beckermr:
Refactor committers collection to include both author and committer logins.

### Description

Closes #1120

### DIFF
--- a/conda_forge_webservices/github_actions_integration/automerge.py
+++ b/conda_forge_webservices/github_actions_integration/automerge.py
@@ -382,7 +382,11 @@ like to enable automerge again!
             return False, "PR does not have the '[bot-automerge]' slug in the title"
 
         # only if only ALLOWED_USERS have commits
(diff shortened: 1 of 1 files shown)


h-vetinari (approved):
Thanks! :)

PR #1319, opened 2026-04-18
Use `ThreadPoolExecutor` for upload pools

jaimergp:
### Description

There's a chance we don't need processes for the upload/copy queue


beckermr:
This appears to work!

PR #1331, opened 2026-04-24
fix: handle missing extra.recipe-maintainers section in recipes for team updates

beckermr:
### Description

Some folks have simply deleted this entire section from their recipe and that causes the webservices to choke. We can handle it if it is missing.


beckermr:
pre-commit.ci autofix

PR #1376, opened 2026-06-05
perf: use "fire and forget" w/ completed futures

beckermr:
### Description

This PR tries to increase the webserver performance by unblocking the main event loop w/ more "fire and forget" and less awaiting for return values.

### DIFF
--- a/conda_forge_webservices/webapp.py
+++ b/conda_forge_webservices/webapp.py
@@ -517,8 +517,9 @@ class UpdateFeedstockHookHandler(WriteErrorAsJSONRequestHandler):
                 return
         else:
             LOGGER.info(f'Unhandled event "{event}".')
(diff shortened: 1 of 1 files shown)


conda_forge_webservices/webapp.py
```diff
@@ -694,13 +722,13 @@ async def post(self):
...
+                tornado.ioloop.IOLoop.current().add_future(fut, _complete_future)
                 return
 
         else:
             LOGGER.info(f'Unhandled event "{event}".')
-
-        self.set_status(404)
-        self.write_error(404)
+            self.set_status(404)
+            self.write_error(404)
```
beckermr:
```suggestion
        self.set_status(404)
        self.write_error(404)
```

PR #1441, opened 2026-08-14
Reply to invalid bot commands in PR comments

### LINKED ISSUE #1429: Reply to incorrect commands
FWICS giving `@conda-forge-admin` an invalid command doesn't trigger any response. This can be quite frustrating, given you don't know whether you did something wrong (I keep mistyping "rerun ci" instead of "restart ci")  or if webservices didn't fire for some reason. I think it would be helpful if the bot replied somehow (either explicitly or via a reaction).

xref #612

mgorny:
### Description

Reply with a comment to apparent webservice requests that do not list any valid command. This helps people realize they've mistyped the command rather than having them wonder whether there is a problem with webservices right now.

So far covered just the PR comment branch. I don't exactly love my code, so I'd appreciate your suggestions on how to make this better.

Fixes #1429

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -213,14 +213,18 @@ def pr_comment(org_name, repo_name, issue_num, comment, comment_id=None):
 def reply_no_command(
     org_name,
     repo_name,
-    pr_num,
+    issue_num,
     comment_id,
     review_id,
 ):
-    if comment_id is not None:
-        request = f"[request](https://github.com/{org_name}/{repo_name}/pull/{pr_num}#issuecomment-{comment_id})"
+    if comment_id == -1:
+        request = (
+            f"[request](https://github.com/{org_name}/{repo_name}/pull/{issue_num}#)"
+        )
+    elif comment_id is not None:
+        request = f"[request](https://github.com/{org_name}/{repo_name}/pull/{issue_num}#issuecomment-{comment_id})"
     elif review_id is not None:
-        request = f"[request](https://github.com/{org_name}/{repo_name}/pull/{pr_num}#discussion_r{review_id})"
+        request = f"[request](https://github.com/{org_name}/{repo_name}/pull/{issue_num}#discussion_r{review_id})"
     else:
         request = "request"
     no_command_message = textwrap.dedent(f"""
@@ -231,12 +235,13 @@ def reply_no_command(
     gh = get_gh_client()
     repo = gh.get_repo(f"{org_name}/{repo_name}")
     if comment_id is not None or review_id is not None:
-        add_reaction("confused", repo, pr_num, comment_id, review_id)
-    pull = repo.get_pull(int(pr_num))
+        add_reaction("confused", repo, issue_num, comment_id, review_id)
     if review_id is not None:
+        pull = repo.get_pull(int(issue_num))
         pull.create_review_comment_reply(review_id, no_command_message)
     else:
-        pull.create_issue_comment(no_command_message)
+        issue = repo.get_issue(int(issue_num))
+        issue.create_comment(no_command_message)
 
 
 def pr_detailed_comment(
@@ -481,6 +486,7 @@ def issue_comment(org_name, repo_name, issue_num, title, comment, comment_id=Non
     ]
 
     if not any(command.search(text) for command in issue_commands):
+        reply_no_command(org_name, repo_name, issue_num, comment_id, None)
         return
 
     # sometimes the webhook outpaces other bits of the API so we try a bit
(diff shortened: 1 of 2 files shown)


conda_forge_webservices/commands.py
```diff
@@ -295,13 +305,23 @@ def pr_detailed_comment(
...
     pr_commands = [LINT_MSG]
     if not is_staged_recipes:
         pr_commands += [ADD_NOARCH_MSG, RERENDER_MSG, UPDATE_CB3_MSG]
 
     if not any(command.search(comment) for command in pr_commands):
+        if not command_found:
+            gh = get_gh_client()
+            repo = gh.get_repo(f"{org_name}/{repo_name}")
+            pull = repo.get_pull(int(pr_num))
+            pull.create_issue_comment(no_command_message)
```
jaimergp: Maybe worth its own little function, since it's repeated twice and I can totally see us failing to update one of the copies in the future when we change it a bit (e.g. to add a "did you mean...?" hint)

conda_forge_webservices/commands.py
```diff
@@ -219,10 +219,17 @@ def pr_detailed_comment(
     comment_id=None,
     review_id=None,
 ):
+    no_command_message = textwrap.dedent("""
+        Hi! This is the friendly automated conda-forge-webservice.
+
+        I couldn't find any valid commands in the request. Please see [Admin web services](https://conda-forge.org/docs/maintainer/infrastructure/#admin-web-services) for the list of valid commands.
+        """)  # ruff: ignore[line-too-long]
```
jaimergp: Can we add a link to the originating message id and tag the author? Sometimes it's tricky to know where it came from (e.g. a comment in a PR review thread, maybe we want to post it in the same thread instead of using the "top-level" discussion).

conda_forge_webservices/commands.py
```diff
@@ -295,13 +305,23 @@ def pr_detailed_comment(
     # below here we only allow staged recipes + feedstocks
     is_staged_recipes = repo_name == "staged-recipes"
     if not (repo_name.endswith("-feedstock") or is_staged_recipes):
+        if not command_found:
+            gh = get_gh_client()
+            repo = gh.get_repo(f"{org_name}/{repo_name}")
+            pull = repo.get_pull(int(pr_num))
+            pull.create_issue_comment(no_command_message)
```
jaimergp: Can we also add a 😞 reaction to the trigger message, in case the issue comment fails for some reason (flaky network, etc)?

conda_forge_webservices/commands.py
mgorny: I didn't even realize review comments were taken into consideration. That said, it will be nontrivial.

mgorny: Ok, done for the "link to the originating message" part. If I am to tag the author, I'd probably need to issue another API call to get them; not sure if we want to add that.

mgorny:
A demo: https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/pull/1790#issuecomment-5302096100

mgorny:
I wish testing this wouldn't have changed my git user though xD.

beckermr:
The link doesn't appear to send me to a list of bot commands in the docs.

mgorny:
> The link doesn't appear to send me to a list of bot commands in the docs.

Are you saying that https://conda-forge.org/docs/maintainer/infrastructure/#admin-web-services doesn't work for you somehow or that this is not a good doc to link at? To be honest, I'd love something better but this is the best I was able to find.

mgorny:
I think I'm done here. Issue, pull requests, comments on both and reviews are now handled. I've also added some docstrings, so people don't have to figure out how the functions are used in the future.

beckermr:
On my phone that link was funky, but I think this was likely my phone. My browser works fine.

beckermr (approved):
Lgtm. I'd not worry about code purity here. The webserver is a bunch of hacks and added up code additions from decades of maintenance. If it works, let's ship it and move.

beckermr:
I am going to merge this into a branch on the upstream repo so I can run the test suite and integration tests.

mgorny:
Thanks! I just wanted to say, I'm not sure if I've tested it right; besides the added tests I ran `pr_detailed_comment()` and `issue_comment()` directly. I did not try to set up the whole server.

Also one thing that's confusing to me is that `COMMAND_PREFIX` filtering occurs only in `pr_comment()` and not the two other functions. I've assumed here that the filtering is now done elsewhere in the stack, and only relevant events reach the server.

beckermr:
For features like this I usually test live. They are non-essential and fully testing in an example webserver is a lot of work for little gain. We do need to run the integration tests of core functions to make sure we don't break the world.

PR #1451, opened 2026-08-20
testing for PR #1441

beckermr:
@mgorny when running live invalid commands cause an infinite loop since the bot comment triggers another webhook event which causes another bot comment etc.

This happens when the invalid command is in the title.

The code needs to be adjusted to search for previous invalid command comments and not make one if one already exists.

### DIFF
--- a/conda_forge_webservices/commands.py
+++ b/conda_forge_webservices/commands.py
@@ -190,7 +190,7 @@ def _attempt_git_clone(repo_url, feedstock_dir, pr_branch=None):
     return repo
 
 
-def pr_comment(org_name, repo_name, issue_num, comment, comment_id=None):
+def pr_comment(org_name, repo_name, issue_num, comment, comment_id=None, actor=None):
     """Process a pull request comment"""
 
     if not COMMAND_PREFIX.search(comment):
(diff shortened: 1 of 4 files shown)


mgorny:
Okay, I don't understand the flow of events then. Does that mean that the comment left by bot is processed by the bot? Why would that happen, given no `@conda-forge-admin` etc. prefix? Doesn't that trigger extraneous commands?

Or do you mean that the bot leaving a comment causes the previous comment to be processed again? But again, doesn't that mean that the same command could be triggered twice?

beckermr:
The bot processes the title of the pr or issue in addition to the comment. So a bad command in a title causes a loop.

mgorny:
> The code needs to be adjusted to search for previous invalid command comments and not make one if one already exists.

This isn't going to work well. Imagine the following scenario:

1. User makes a comment with invalid command.
2. Bot replies.
3. User tries again, makes another comment, the command is still invalid though.
4. Bot doesn't reply anymore.

Detecting existing comment could work _if_ we detected the comment referring to the exact same request. However, I find it hard to work on this without seeing the actual calls / events. Any chance you could paste them in some form here?

beckermr:
I don't have a sequence of events in one spot.

beckermr:
The other option is to detect who made the comment and to not respond to certain bot comments.

beckermr:
I think for now we could merge this one and then fix this up later if you wanted.

PR #1493, opened 2026-09-09
feat: set status on commits for automerge

beckermr:
### Description

This will help us debug problems.

cc @h-vetinari

### DIFF
--- a/conda_forge_webservices/github_actions_integration/__main__.py
+++ b/conda_forge_webservices/github_actions_integration/__main__.py
@@ -11,7 +11,7 @@ from conda_forge_feedstock_ops import setup_logging
 from conda_forge_feedstock_ops.lint import lint as lint_feedstock
 from git import Repo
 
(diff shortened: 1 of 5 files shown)


h-vetinari:
Thanks for this!

Can we make the status conditional on whether the automerge label has been set? For now it just seems to be generating a failing check (e.g. in https://github.com/conda-forge/clangdev-feedstock/pull/491) due to
```
INFO:conda_forge_webservices.github_actions_integration.automerge:DID NOT MERGE PR 491 on conda-forge/clangdev-feedstock: user h-vetinari cannot automerge and no automerge label found
```

h-vetinari:
I think we may also need to be careful that a failing automerge check isn't the only thing (assuming all other checks are green) that causes automerge to fail 🤣

beckermr:
Automerge ignores it own checks for sure. I can make the status conditional on a label.

PR #1509, opened 2026-09-17
Refactor conda-forge-admin PR helpers

chrisburr:
### Description

Refactor conda-forge-admin PR helpers in preparation for using trusted publishing in conda-forge/release.


.github/workflows/tests.yml
```diff
@@ -356,7 +356,7 @@ jobs:
           export CF_FEEDSTOCK_OPS_CONTAINER_TAG="${version}"
 
           cd tests
-          pytest -vvs --branch=${branch} test_live_rerender.py
+          pytest -vvs --branch=${branch} test_live_rerender.py test_live_admin_pr.py
```
beckermr: This will run this test in serial with the rerender test, where as I think it can be run in parallel. Testing time has been an issue here. I'd rather copy-and-paste a new CI job like the others. The other option would be to use pytest-xdist with `n=2`.

beckermr:
Thanks for the fix! I have added the PR to the merge queue. It will run the integration tests there. If they pass, the new changes will get deployed.

beckermr:
Looks like the admin pr tests failed because it needed the CF webservices credentials (PEM and app id).

beckermr:
link: https://github.com/conda-forge/conda-forge-webservices/actions/runs/35262921758/job/105344331046

PR #1512, opened 2026-09-18
feat: verify OIDC identity tokens from CI providers for trusted publishing

chrisburr:


beckermr:
Wow, I was right that there is no way I could personally implement this securely. Kudos!

beckermr:
It looks like most of the error cases are covered in the tests. I assume we can add a test against a live repo once the actual endpoints are done?

Also, it'd be good in the live repo tests to send bad requests and make sure they are rejected at runtime. 

The current upload tests show how to bootup the webserver in the CI so you can test against it.

chrisburr:
yeah absolutely. I thought it was good to get the crypto side of things first, given it fits nicely into a standalone PR and couples to the conda smithy one

conda_forge_webservices/trusted_issuers.yaml
```diff
@@ -0,0 +1,20 @@
+# The OIDC issuers whose tokens conda-forge accepts for trusted publishing.
+# A feedstock's trusted_publishers can only name an issuer listed here, and
+# deleting an entry stops its tokens being accepted from the next deploy.
+#
+#   issuer    the token's `iss`, exactly as the issuer writes it
+#   provider  the kind of trusted_publishers entry that may name it
+#   jwks_uri  where the issuer publishes its signing keys, as its
```
chrisburr: I ended up deciding to let the list of supported issuers and URLs for getting public keys to be hard coded rather than using OIDC. Otherwise, there are too many unlikely-but-problematic risks with malicious servers which become extremely tedious to defend against (e.g. redirecting to localhost of the heroku server).

Hard coding the `jwks_uri` negates the entire space of threats at the cost of maybe needing to update the config now and again.

conda_forge_webservices/trusted_publishing.py
```diff
@@ -0,0 +1,687 @@
...
+
+REQUIRED_CLAIMS = ["iss", "aud", "exp", "iat", "sub", "jti"]
+
+# a runner's clock and ours will not agree to the second
+LEEWAY = 30
+
+# How long after it was issued a token is taken, whatever its exp says.
+# GitLab's last as long as the job's timeout, an hour by default, and one
+# that leaks before it is used would otherwise be good for all of that.
+MAX_TOKEN_AGE = 15 * 60
```
chrisburr: This is more restrictive than PyPI does for their trusted publishing but given GitLab defaults to long lifetimes I decided to lock things down further. We can always relax this constraint if it becomes annoying.

conda_forge_webservices/trusted_issuers.yaml
```diff
@@ -0,0 +1,20 @@
+# The OIDC issuers whose tokens conda-forge accepts for trusted publishing.
```
chrisburr: I decided to put this file in here to make it easier to add/remove issuers without needing a smithy release.

We can fairly freely add hosts here but the presence of it makes it much easier to reject invalid requests without making requests to GitHub or the IdP.

conda_forge_webservices/trusted_publishing.py
```diff
@@ -0,0 +1,687 @@
...
+    if issuer == GITHUB_ISSUER:
+        return f"{claims.get('job_workflow_ref')} (run {claims.get('run_id')})"
+
+    host = issuer.removeprefix("https://")
+    return f"{host}/{claims.get('project_path')} " + (
+        f"({claims.get('ref_type')} {claims.get('ref')})"
+    )
+
+
+# The tokens authorize has accepted, kept until they expire so that one that
```
chrisburr: This follows how PyPI does things and I see no reason not to. We loose the invalidation on restarts but I think that's fine give the lack of a good place to synchronise using.

conda_forge_webservices/trusted_publishing.py
```diff
@@ -0,0 +1,687 @@
+"""Verify the identity tokens that CI providers issue to their own jobs.
```
chrisburr: A lot of this is arguably overkill given it will only be able to update a feedstock to a real version so there is very little value even if a token does leak.

That said, I could see this being useful in other places where we talk to the web-services from GitHub so there is no reason not to do it as best I can.

